Issue #7, opened 2019-02-22
Security context not respected


unguiculus:
I'm trying to use `local-path-provisioner` with [kind](https://github.com/kubernetes-sigs/kind). While it seems to generally work with multi-node clusters, security contexts are not respected. Volumes are always mounted with `root` as group. Here's a simple example that demonstrates this:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: local-path-test
  labels:
    app.kubernetes.io/name: local-path-test
spec:
  containers:
    - name: test
      image: busybox
      command:
        - /config/test.sh
      volumeMounts:
        - name: test
          mountPath: /test
        - name: config
          mountPath: /config
  securityContext:
    fsGroup: 1000
    runAsNonRoot: true
    runAsUser: 1000
  terminationGracePeriodSeconds: 0
  volumes:
    - name: test
      persistentVolumeClaim:
        claimName: local-path-test
    - name: config
      configMap:
        name: local-path-test
        defaultMode: 0555

---
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: local-path-test
  labels:
    app.kubernetes.io/name: local-path-test
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: "1Gi"
  storageClassName: local-path

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: local-path-test
  labels:
    app.kubernetes.io/name: local-path-test
data:
  test.sh: |
    #!/bin/sh

    ls -al /test

    echo 'Hello from local-path-test'
    cp /config/text.txt /test/test.txt
    touch /test/foo

  text.txt: |
    some test content
```

Here's the log from the container:

```
total 4
drwxr-xr-x    2 root     root            40 Feb 22 09:50 .
drwxr-xr-x    1 root     root          4096 Feb 22 09:50 ..
Hello from local-path-test
cp: can't create '/test/test.txt': Permission denied
touch: /test/foo: Permission denied
```

As can be seen, the mounted volume has `root` as group instead of `1000` as specified by the security context. I also installed `local-path-provisioner` on Docker4Mac. The result is the same, so it is not a `kind` issue. Using the default storage class on Docker4Mac, it works as expected.

rimusz:
yes, I can confirm that too!

aojea:
[...]

PR #9, opened 2019-02-27
Fix permissions

unguiculus:
This is a follow-up PR for #7, since https://github.com/rancher/local-path-provisioner/commit/3256bdf61b14c6b7e517ede0f56cfebeaf28e616, unfortunately, did not fully fix the issue.

The example I supplied is not ideal. It works with your fix because it uses Busybox which doesn't have group 1000 so the user runs with gid=0. I added `id` to the script which prints the following:

`uid=1000 gid=0(root) groups=0(root),1000`

Now, when I use an image that actually has a user with uid=1000, gid=1000, such as `jenkins/jenkins` it doesn't work because the user does not belong to group 0 in this case:

`uid=1000(jenkins) gid=1000(jenkins) groups=1000(jenkins)`

This fixes the issue using `0777` as Minikube does.

### DIFF
--- a/examples/pod-with-security-context.yaml
+++ b/examples/pod-with-security-context.yaml
@@ -7,7 +7,9 @@ metadata:
 spec:
   containers:
     - name: test
-      image: busybox
+      # Jenkins image used for illustration purposes because it has a user
+      # with uid=1000 and gid=1000 that matches the specified security context
+      image: jenkins/jenkins
       command:
         - /config/test.sh
       volumeMounts:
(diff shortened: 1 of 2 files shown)


yasker:
@unguiculus I see.

`0777` won't be very secure, but I cannot see a better solution for now since we don't have any information regarding what the permission should be in [`opts`](https://github.com/rancher/local-path-provisioner/blob/master/provisioner.go#L164)when creating the volume.

Can you update the [example](https://github.com/rancher/local-path-provisioner/blob/master/examples/pod-with-security-context.yaml) as well? I will merge and release a new version after.

yasker:
@unguiculus Sorry, didn't see your update sooner. Github still doesn't send notification for updated commit..

Merged, thanks.

yasker:
@unguiculus Local Path Provisioner v0.0.4 released.

PR #11, opened 2019-03-14
use the correct string for namespace env

kmova:
When deploying this provisioner in non-default namespace, this provisioning fails with an message about missing local-path-storage. 

As per the deployment file, Namespace is passed using the string POD_NAMESPACE.  (https://github.com/rancher/local-path-provisioner/blob/master/deploy/local-path-storage.yaml#L75)

This PR is to correct the variable name used for fetching the namespace.


yasker:
Nice catch! Merged, thanks @kmova

PR #13, opened 2019-04-09
build and packaging updates

anitgandhi:
## Changes

1. Update Dockerfile.dapper to install Go 1.12, separate go gets for better image layer caching
2. replace ubuntu image with simple alpine base image
3. add linker flag -w for smaller binary
4. gitignore some generated files

I've tested this in a Vagrant box and it seems to work fine. Since the binary was built with CGO_ENABLED=0 already there was no need for anything else at runtime. I initially tested with `FROM scratch` but that failed since `/tmp` didn't exist (for logs), so I switched to just plain `alpine:latest`.

The Docker image goes down from ~320MB (v0.0.5) to ~30MB :smile:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
+*dapper*
+!Dockerfile.dapper
+bin/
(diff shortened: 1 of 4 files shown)


yasker:
@anitgandhi It looks good! Can you merge two commits into one? Since the change to Alpine was originated from the first commit.

anitgandhi:
@yasker done!

yasker:
Merged, thanks!

@anitgandhi Since it's not a blocker issue, I will release the new version later. Is that OK?

anitgandhi:
@yasker yup no problem!

anitgandhi:
@yasker actually, do you think it would be possible to make a v0.0.6 with these changes? I know there's no functional changes, but it would make things a bit easier for us if we can include these changes :)

yasker:
@anitgandhi No problem. v0.0.6 has been cut. Let me know if it works for you.

Issue #12, opened 2019-03-17
Add arm64v8 image


vchrisb:
adding arm64v8 support would be great!

giminni:
I installed successfully k3s on arm64 with weave Net overlay network on a RockPro64 cluster 
and made my own private local-path-provisioner arm64v8 instance using Alpine Linux instead of Ubuntu:18.04. Alpine Linux image size is 84Mi, testing was successful. 

```
$ uname -a
Linux <server> 4.4.167-1161-rockchip-ayufan-g6f1664023387 #1 SMP Fri Mar 22 23:03:38 UTC 2019 aarch64 aarch64 aarch64 GNU/Linux
<user>@<server>:<home-dir>$ k get all -n local-path-storage
NAME                                          READY   STATUS    RESTARTS   AGE
pod/local-path-provisioner-6868fbd6c6-nvjk5   1/1     Running   0          22m

NAME                                     READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/local-path-provisioner   1/1     1            1           22m

NAME                                                DESIRED   CURRENT   READY   AGE
replicaset.apps/local-path-provisioner-6868fbd6c6   1         1         1       22m
```
IMHO alpine linux reduce the footprint even further.

Should I make a PR using Alpine Linux?

yasker:
@giminni Sure, that would be great!

Though I don't have a setup to test arm64, so I would be unable to fix if there is anything wrong with Arm. But I am OK on accepting a PR for arm64 as long as there is no impact on running with x86.

anitgandhi:
@giminni we switched to Alpine in #13 , I think adding arm64/multi-arch support can build off that potentially?

optimuspaul:
have any arm images been pushed anywhere?

yasker:
Can anyone try @km4rcus 's image? The build has been merged to master as well.

vchrisb:
@yasker can you initiate a build to docker hub?

optimuspaul:
I haven't tried that specific build but I did build it for arm myself and it is working

yasker:
v0.0.8 released and the image pushed. Let me know if it works.

vchrisb:
@yasker I don't see any multi-arch image tags on docker hub and also no manifest to support multi-arch

yasker:
@vchrisb sorry, not familiar with multi arch build. Let me check what should I do for an arm image.

tamsky:
[...]

PR #19, opened 2019-05-01
Add arm64 support

km4rcus:
This PR should address #12. 
I have built an arm64 docker image km4rcus/local-path-provisioner:arm64


yasker:
Merged, thanks!

I cannot test this image, so I would wait for a while for others to confirm.

PR #24, opened 2019-05-28
working: multi-arch-images

tamsky:
This has been a very interesting area of docker to learn about.
Hopefully this helps folks get more things running on ARM and other platforms.

I copied the basic multi-arch build structure from https://github.com/rancher/dapper 
along with a couple of their `scripts/*`.

I'm not certain the `go build ... -o binary.arch` suffixes in the `scripts/build` I included in this PR are 100% correct.

I chose these so they match the `arch`+`variant` names associated with the current `alpine` manifest,  (read: not the same as the bin_arch names used in https://github.com/rancher/dapper/blob/master/scripts/build .)

Also,
It looked like the `multi-arch-images` `make` target in `rancher/dapper` was being manually called -- as well as the resulting `push.sh` it emitted... so I'm not sure where/how the CI pipelines build and push the multi-arch images... ?

I'm not sure if `dapper` handles exposing files created in containers back to the host, but there is now a `manifest.yaml` file that gets created and should be fed to the `push.sh` script which invokes `manifest-tool`.

If there's a better way, please let me know, or maintainers can push more commits to this branch.

Relates to #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,28 @@ To uninstall, execute:
 kubectl delete -f https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml
 ```
 
+## Building from source
+
+To build binaries for all supported architectures
+```
+$ CROSS=1 make
+```
+The compiled binaries are written to `./bin` with filenames like `./bin/local-path-provisioner-linux-armv6`.
+
+You will need a locally installed copy of the `manifest-tool`
+https://github.com/estesp/manifest-tool, in order to push the
+resulting docker images.
+
+To process the compiled binaries into per-arch and assemble the final multi-arch image, run:
+```
+$ make multi-arch-images
+```
+
+At the end of the `make` output are two inline files, `manifest.yaml`
+and `push.sh`.  Copy each of them from the stderr/stdout to their
+respective filenames, and then run `source push.sh` to push the
+individual arch images and create the final multi-arch image.
+
 ## License
 
 Copyright (c) 2014-2018  [Rancher Labs, Inc.](http://rancher.com/)
--- a/scripts/ci
+++ b/scripts/ci
@@ -7,8 +7,3 @@ cd $(dirname $0)
 ./validate
 ./test -cover
 ./package
-
-image=`cat ../bin/latest_image`
-
-echo
-echo Longhorn Manager image: ${image}
--- a/scripts/multi-arch-images
+++ b/scripts/multi-arch-images
@@ -1,0 +1,81 @@
+#!/bin/bash
+set -ex
+
+IMAGES=
+REPO=${REPO:-rancher}
+
+cd $(dirname $0)/..
+
+. ./scripts/version
+VERSION="${VER}"  # apparently, from rancher repo to repo, this script can't pick a common output var
+
+# create an ephemeral push script
+echo '#!/bin/bash' >push.sh
+chmod +x push.sh
+
+# create an ephemeral manifest file
+echo image: ${REPO}/local-path-provisioner:${VERSION} > manifest.yaml
+echo manifests: >> manifest.yaml
+
+while read HASH OS VARIANT ARCH; do
+    BIN_ARCH=${ARCH}
+    case ${OS}-${BIN_ARCH} in
+        Darwin-x86_64)
+            ;;
+        linux-arm)
+            [[ ${VARIANT} == v6 ]] && BIN_ARCH=armv6
+            [[ ${VARIANT} == v7 ]] && BIN_ARCH=armv7
+            ;;
+        linux-amd64)
+            ;;
+        linux-arm64)
+            VARIANT=null
+            ;;
+        linux-386|linux-ppc64le|linux-s390x)
+            continue ;;  # skips these ARCH.  Are these platforms supported yet?
+        *)
+            echo ARCH $BIN_ARCH not found.  This is a problem.
+            exit 1
+            ;;
+    esac
+
+    cat > Dockerfile.gen << EOF
+FROM alpine@${HASH}
+COPY bin/local-path-provisioner-${OS}-${BIN_ARCH} /usr/local/bin/local-path-provisioner
+EOF
+    docker build -f Dockerfile.gen -t ${REPO}/local-path-provisioner:${VERSION}-${BIN_ARCH} .
+
+    # add to list of images to push
+    echo docker push ${REPO}/local-path-provisioner:${VERSION}-${BIN_ARCH} >> push.sh
+
+    # handle lack of useful VARIANT template handler when VARIANT is null (jq emits 'null' for us),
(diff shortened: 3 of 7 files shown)


tamsky:
I've pushed images from this branch to https://cloud.docker.com/repository/registry-1.docker.io/tamsky/local-path-provisioner/tags

Once pushed to docker hub, multi-arch enabled images list the supported architectures when you hover over the penguin:

image

yasker:
Thanks @tamsky . I would need to take some time to look into the details. But please squash the commit into one first (since it's doing one thing, which is adding multi-arch images support)

tamsky:
> I would need to take some time to look into the details. 

Rather than expecting every member of the Rancher Org to be an expert at all things, may I suggest reaching out to any others in your Org that have already done this type of work for review assistance?  Perhaps the folks that added multi-arch to `rancher/dapper` could help, since I borrowed a lot from their script.

> But please squash the commit into one first 

Also, since I don't believe this PR is finished yet, can we wait until after reviews are complete before squashing (as a last step)?

Lastly, in case you were not aware of it github's "Merge pull request" button now allows you to choose one of these merge strategies:
- Create a merge commit
- Squash and merge
- Rebase and merge

cryptk:
For what it's worth, I tested this last night on a Raspberry Pi 4 running k3s and it seemed to work quite well using the images from https://hub.docker.com/r/tamsky/local-path-provisioner.  The RPi4 is an armv7, so I can at least vouch for that image working

scripts/ci
```diff
@@ -7,8 +7,3 @@ cd $(dirname $0)
 ./validate
 ./test -cover
 ./package
-
```
yasker: Why remove those?

README.md
```diff
@@ -172,6 +172,23 @@ To uninstall, execute:
...
 ```
 
+## Building from source
+
+To build binaries for all supported architectures
+```
+$ CROSS=1 make
+```
+
+To build docker images from those binaries, you will need a local copy of the `manifest-tool` installed.
```
yasker: I saw this from https://github.com/estesp/manifest-tool:

>UPDATE (Feb 2018): The Docker client PR #138 is merged, providing a docker manifest command that can replace the use of manifest-tool in most scenarios. Other follow-on PRs are in process to add functionality to docker manifest to make it completely functional for all multi-platform image use cases.

Should we use that instead?

scripts/package
```diff
@@ -3,24 +3,5 @@ set -e
...
-
-cp -a ./bin ./package/
-trap 'rm -rf ./package/bin' exit
-docker build -t ${IMAGE} ./package
-
-echo Built ${IMAGE}
-
-echo ${IMAGE} > ./bin/latest_image
+mkdir -p dist/artifacts
+cp -r bin/local-path-provisioner* dist/artifacts
```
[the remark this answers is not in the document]
yasker: `scripts/package` generates the docker image.

scripts/multi-arch-images
```diff
@@ -0,0 +1,81 @@
...
+            ;;
+    esac >> manifest.yaml
+
+done <<< $(DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect alpine:3.9 | jq -r '.manifests[] | "\(.digest) \(.platform.os) \(.platform.variant) \(.platform.architecture)"')
+
+echo '#######' manifest.yaml
+cat ./manifest.yaml
+echo '#######' manifest.yaml
+
+echo manifest-tool push from-spec manifest.yaml >> push.sh
```
yasker: Pushing is separated from building the image (which is handled by `scripts/package`). It shouldn't be automated. Push steps can be documented instead.

README.md
```diff
@@ -187,7 +187,7 @@ https://github.com/estesp/manifest-tool
 $ make multi-arch-images
 ```
 
-
+At the end of the output are two files, `manifest.yaml` and `push.sh`, copy them from the stderr/stdout to their respective filenames, and then run `source push.sh` to push the individual arch images and create the multi-arch image.
```
yasker: Merge with the previous commit.

scripts/multi-arch-images
tamsky: > `scripts/package` generates the docker image.

Yes, but in this PR, that responsibility will move to the new script: `scripts/multi-arch-images`.

This is exactly the same the layout used by:
- https://github.com/rancher/dapper/blob/master/scripts/package
- https://github.com/rancher/dapper/blob/master/scripts/multi-arch-images

Is there some reason we are unable to utilize that project's code and structure?

tamsky: > Pushing is separated from building the image (which is handled by `scripts/package`). It shouldn't be automated. Push steps can be documented instead.

I can comment out the line where `push.sh` is executed.
Would that be acceptable?

scripts/ci
tamsky: > Why remove those?

Because:
- `./bin/latest_image` is no longer being created by the `scripts/package` script.
- `latest_image` no longer appears in this repo.
- A missing file will cause this script to fail.
- Finally, if these `echo` statements are actually important, a comment should be present in this file explaining their importance.

tamsky:
Squashed to one commit, as requested.

README.md
tamsky: After this PR is accepted, I would consider it.

tamsky: I would also expect that same new PR to be one that `rancher/dapper` would also be interested in using, so I will open similar PRs together, at that time.

cryptk:
Any word on if this will get merged or not? I would love to be able to use official images on my RPi4 k3s cluster

yasker:
@cryptk Yeah, I am working on this. Sorry for the delay, I've been busy recently with Longhorn.

yasker:
Hi @tamsky 

Sorry for the delay... Been too busy with Longhorn.

After reading into the PR more, I realized there are multiple steps that lead to build a multiarch image, so I created PR #42 to cover the first two steps: build multiarch binary and update docker to support `manifest` command. Can you give it a check to see if it looks OK?

And obviously the next step is:
1. Build images for each architecture
2. Generate a multiarch image using all the previously built images (it's better if we can use `docker manifest` all the time)

And I want this step to follow the previous pattern, e.g. keep the `scripts/package` for the packaging purpose. It can be similar to what `scripts/build` do, use `CROSS=1` env to trigger multiarch image build instead of host image build.

I've added a script `script/release` with the intent to generate multiarch build.

Another thing is, we won't want to push automatically - since we don't want to push the intermediate image to the official repo. So it's better even for the `release` script, we just show the steps needed to push the image, but not really push the image.

And we can still keep the `latest_image` generated since we can use that string to update the `/deploy/` (yeah, I probably want to add a script for that).

Thanks again for contribution. Sorry for the delay.

yasker:
Close this PR since @ibuildthecloud has added the multiarch image support.

PR #32, opened 2019-07-09
support deployment on air-gap

dorsany:
on air-gap mode if you use ```imagePullPolicy: Always``` kubernetes will try to downlod the image even there is no any internet connection. therefor if we will use ```imagePullPolicy: IfNotPresent``` if the image was pre-loaded it will work

please consider to confirm this PR to support air-gap


yasker:
@dorsany Merged, thanks.

PR #45, opened 2019-09-20
Remove namespaces from cluster scoped objects

ibuildthecloud:
Related to this bug in wrangler https://github.com/rancher/wrangler/pull/40


yasker:
Thanks for catching it! @ibuildthecloud

PR #66, opened 2019-12-24
pretty storage path & github actions

egeneralov:
Hello!
I suggesting use Github Actions for releases and human-readable names of PV, like `${BASEDIR}/${NAMESPACE}/${PVC-NAME}`.


runningman84:
Would you mind adding a pre command (should be called before doing the mkdir)? This would allow us to use btrfs subolumes...

egeneralov:
@runningman84
You mean `btrfs subvolume create ${PATH}`? I think, it's possible, and can be added as additional feature. Will do this at week.

runningman84:
yes that would be great

yasker:
@egeneralov Thanks for the PR!

1. I haven't used GitHub Action yet but this looks great. However, previously we can build and run validation using `make` command locally (e.g. `go fmt`, `go lint`). I would love to still have that ability to verify the build locally before pushing to the repo.
2. Use `${NAMESPACE}/${PVC-NAME}` might be a bit confusing since PV can be reused by a different PVC (though we've set `reclaimPolicy` to `Delete` by default). Also, the deletion part will break the backward compatibility since the naming pattern changed. I am a bit hesitant to accept this change due to those reasons.

runningman84:
@yasker what would you think about my idea with a pre command (for creation) and maybe a post command (for deletion) in order to support btrfs subvolumes?

yasker:
@runningman84 Yeah, we can do that. It can be done by adding more options in [our configuration](https://github.com/rancher/local-path-provisioner#configuration). We would need a bit design on this, e.g. do we want to have it configured globally or per node?

runningman84:
I would like to have it here:
```
                {
                        "node":"DEFAULT_PATH_FOR_NON_LISTED_NODES",
                        "paths":["/opt/local-path-provisioner"]
                        "create_cmd":"btrfs subvolume create ${PATH}",
                        "delete_cmd":"btrfs subvolume delete ${PATH}"
                },
                {
                        "node":"yasker-lp-dev1",
                        "paths":["/opt/local-path-provisioner", "/data1"]
                        "create_cmd":"lvm create blabla ${PATH}",
                        "delete_cmd":"lvm delete blabla ${PATH}",
                        "reszie_cmd":"lvm resize"
                },
                {
                        "node":"yasker-lp-dev3",
                        "paths":[]
                        "create_cmd":null,
                        "delete_cmd":null
                }
```
It would be greate to have examples for btrfs and lvm which are most commonly used... with variables for PVC name, storage size and so on we could also ensure that no single volume would max out a single node.

yasker:
@runningman84 Those look good in general. Can you file a feature request? Don't want to hijack the thread of PR.

runningman84:
done https://github.com/rancher/local-path-provisioner/issues/67

bk201:
@egeneralov  Hi, are you still able to work on this PR?
I would suggest rebasing and focus on only GitHub actions.
- For the pretty storage path part, the code was already changed, and changing it again might cause issues after upgrading (deleting a non-exist path).
- Do not removing the scripts folder to be consistent with other Rancher projects.

innobead:
Closing the PR, it seems pending for a while.

@egeneralov feel free to contribute back.

PR #71, opened 2020-01-15
Use sigs.k8s.io/sig-storage-lib-external-provisioner

tamalsaha:
- Uses PV controller from new library location
- Includes fix for k8s 1.11 ref:https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/60

Signed-off-by: Tamal Saha <tamal@appscode.com>


yasker:
Thanks for the PR @tamalsaha !

1. Is there a minimum Kubernetes version requirement for the new library? `go.mod` used v0.17.1, not sure what's the matching Kubernetes version.
2. Can you separate the functional change from the automatic generated change?

tamalsaha:
> Is there a minimum Kubernetes version requirement for the new library? go.mod used v0.17.1, not sure what's the matching Kubernetes version.

This is Kubernetes 1.17.1. https://twitter.com/TheNikhita/status/1204556329458622464

> Can you separate the functional change from the automatic generated change?

From this repo's POV, I have only updated the go.mod file to use the library from its' new location and made a change from VolumeOptions -> ProvisionOptions . https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/46 . There is no functional change in this repo.

yasker:
@tamalsaha 

1. Does that mean we only support v1.17.1+ from now on? Since we cannot break all the users use the previous version or force them to upgrade.
2. What I was concerned about is it's hard to tell apart the code written by hand and the generated code. But now I think it's fine for now since it seems we don't have a clear way to separate them without break the build (or `git bisect`).

tamalsaha:
> Does that mean we only support v1.17.1+ from now on? Since we cannot break all the users use the previous version or force them to upgrade.

No, local-host-provisioner will work for any version from Kubernetes 1.11. We are already using this with [Stash from my custom build](https://github.com/stashed/stash/commit/64f75bb72a0477a04b276a990a7eb0dc63f2c970/checks?check_suite_id=399547899).

yasker:
LGTM. Merged, thanks.

tamalsaha:
Thanks a lot, @yasker! Can you please cut a release? I would like to use the new version in kind.

yasker:
@tamalsaha  Done. Let me know if it works for you.

tamalsaha:
Thank you!

Issue #77, opened 2020-03-01
[FEATURE REQUEST] More obvious folder names


REBELinBLUE:
This is a feature request but I would like to suggest a change to how the folders are named.

When browsing the file system the names are non-obvious, other provisioners, such as the NFS provisioner include the namespace and claim name in the folder name along with the PV name.

So where you have https://github.com/rancher/local-path-provisioner/blob/master/provisioner.go#L183-L184

NFS provisioner has https://github.com/kubernetes-incubator/external-storage/blob/master/nfs-client/cmd/nfs-client-provisioner/provisioner.go#L62-L65

It looks like a simple change and I can do a PR if you agree

yasker:
Sounds good. We need to make sure the existing volumes won't be impacted.

I was against using the PVC name since it's not cluster-wide unique and can cause confusion (even with the namespace) if the PV was reused. But I like the NFS provisioner's approach since it also put the PV name into the path.

So I think something like: `<pv-name>_<pvc-namespace>_<pvc-name>` would be good. It might a bit more trouble if you want to list directory for certain namespace, but I do want to make it obvious that PV is the defining element here, not PVC namespace or PVC name. Let me know what do you think.

REBELinBLUE:
I'm having a play around with it this evening. I don't really know much about StorageClasses and PVs but my understanding is nothing will be needed to make it continue working with existing volumes because the path will be defined in the PV which has already been provisioned won't it?

varac:
Nice to see this feature merged, can you please release a new version so we could profit from this ?

ariep:
Is there a release coming up that will include this?

yasker:
@varac v0.0.13 has been released. Enjoy!

PR #78, opened 2020-03-03
Change to provisioned path to include PVC Namespace and Claim name

REBELinBLUE:
As discussed in #77 

This seems deceptively simple, am I missing something? I can't seem to find any existing tests?

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -180,8 +180,11 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	if err != nil {
 		return nil, err
 	}
+
 	name := opts.PVName
-	path := filepath.Join(basePath, name)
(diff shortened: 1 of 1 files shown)


provisioner.go
```diff
@@ -180,8 +180,11 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	if err != nil {
 		return nil, err
 	}
+
 	name := opts.PVName
-	path := filepath.Join(basePath, name)
+	folderName := strings.Join([]string{name, opts.PVC.Namespace, opts.PVC.Name}, "-")
```
yasker: Use `_` instead of `-`? Make it easier to tell apart different parts of the name.

yasker:
Can you fold/squash the two commits into one commit? Otherwise LGTM.

REBELinBLUE:
Done

yasker:
Merged, thanks @REBELinBLUE

Issue #73, opened 2020-02-05
High CPU Usage


mschae:
See #14 

I am using k3s on a i5 and the local path provisioner (version v0.0.11) sits at 18% CPU:
![grafik](https://user-images.githubusercontent.com/473241/73830577-2f128280-4805-11ea-8139-eec5fd2c1ee2.png)

Steps to reproduce:
1. Install k3s on x86 Hardware
2. Use the local path provisioner

Interestingly enough I am also running k3s on a cloud server, not using the local-path-provisioner (but having it running) and it doesn't consume nearly as many resources there.

mshade:
I'm seeing this as well, with v0.0.11. For me, cpu usage starts low if I delete the pod and let it recreate, then after a while I see it back up in my top consumers.

I've just updated the manifest to v0.0.12 and will see if the issue persists.

PR #87, opened 2020-05-25
Fix time.Tick leak

dittos:
https://golang.org/pkg/time/#Tick

> While Tick is useful for clients that have no need to shut down the Ticker, be aware that without a way to shut it down the underlying Ticker cannot be recovered by the garbage collector; it "leaks".

It seems to fix high CPU usage issue. (#73)

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -119,9 +119,11 @@ func (p *LocalPathProvisioner) refreshConfig() error {
 
 func (p *LocalPathProvisioner) watchAndRefreshConfig() {
 	go func() {
+		ticker := time.NewTicker(ConfigFileCheckInterval)
+		defer ticker.Stop()
 		for {
 			select {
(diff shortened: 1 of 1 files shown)


yasker:
Thanks for the PR @dittos 

I am not sure this can cause the CPU utilization issue though since this code is only called once and will be terminated with the process.

dittos:
> this code is only called once and will be terminated with the process.

Isn't `time.Tick` called every 30 secs in the loop?

I ran the patched version and it fixed the CPU usage increasing.

**left (green): patched version / right (red): v0.0.13**

Screenshot_2020-05-31 Prometheus Time Series Collection and Processing Server

Note that there was a simillar fix in the other project: https://github.com/jetstack/cert-manager/pull/2467

yasker:
@dittos You're right. I misread the code.

Thanks for troubleshooting this! I will merge this and do a new release.

PR #91, opened 2020-06-11
Create local volumes instead of hostPath

### LINKED ISSUE #85: enable velero backups by using local instead of hostpath
by using local instead of hostpath it would be possible to use velero with restic for backups.
velero with restic for volume backups can not support to backup host-path, but local is supported.

i use cassandra with rancher/local-path-provisioner for volumes to get bare metal disk performance, but backups are also nice...

velero limitations
https://github.com/vmware-tanzu/velero/blob/master/site/docs/master/restic.md#limitations

host-path vs local
https://kubernetes.io/docs/concepts/storage/volumes/#local
https://kubernetes.io/docs/concepts/storage/volumes/#hostpath
https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga/

as far as a have read the code, the only place that would need to change is https://github.com/rancher/local-path-provisioner/blob/655eac7962bc1dbafb4bdec60b86bc8bc76b307c/provisioner.go#L212

does a technical reason exist not to use local?
would it break current deployments to have a mixture of volumes types?
is a merge request welcome?

ariep:
Create local volumes instead of hostPath ones. Closes #85 .

This needs some documentation changes at the least.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -199,7 +199,6 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	}
 
 	fs := v1.PersistentVolumeFilesystem
-	hostPathType := v1.HostPathDirectoryOrCreate
 	return &v1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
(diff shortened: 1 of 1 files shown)


yasker:
The code part LGTM. We do need a bit more documentation updates.

Also, I am not using Velero so it would be helpful if someone else can give this PR a test with Velero.

yasker:
I will merge this for now. Let's test it for a few days before releasing the next version.

Issue #95, opened 2020-06-17
add scripts to make it configurable for users when add/removing a dir 


nicktming:
when provisioning a pv, some users might want to add some operations besides (mkdir -p or rm -f), but now it is hardcoded in local-path-provisioner project.

i made some changes to make it configurable for users, i put it in ConfigMap(https://github.com/nicktming/local-path-provisioner/tree/dev-scripts/deploy/cm) when adding or removing a dir, the helper pod read add.sh and del.sh to config cmdsForPath in provisioner.go

if it is helpful, i would like to commit a pr for this feature.

yasker:
@nicktming I like that idea. Yes, we can make them configurable.

There might be more we can do, e.g. set a quota for btrfs if needed if we implement certain parameter replacement. But that would be for later.

nicktming:
@yasker i commit a pr at #99 which makes it configurable for users.

As for quota, i think users may have different requirements, for example users could use xqs_quota command at scripts if they use xfs. but it would be great if this project could set a quota regardless of file systems. I am also working on it and  i prefer to use xqs_quota command in configurable scripts now.

PR #99, opened 2020-06-24
Feature/script configurable

nicktming:
For this issue: #95 

add scripts to make it configurable for users when add/removing a dir.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,14 @@ data:
                 }
                 ]
         }
+  setup: |-
+        #!/bin/sh
+        path=$1
+        mkdir -m 0777 -p ${path}
+  teardown: |-
+        #!/bin/sh
+        path=$1
+        rm -rf ${path}
 
 ```
 
--- a/deploy/local-path-storage.yaml
+++ b/deploy/local-path-storage.yaml
@@ -102,4 +102,12 @@ data:
                 }
                 ]
         }
+  setup: |-
+        #!/bin/sh
+        path=$1
+        mkdir -m 0777 -p ${path}
+  teardown: |-
+        #!/bin/sh
+        path=$1
+        rm -rf ${path}
 
--- a/provisioner.go
+++ b/provisioner.go
@@ -190,9 +190,8 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	logrus.Infof("Creating volume %v at %v:%v", name, node.Name, path)
 
 	createCmdsForPath := []string{
-		"mkdir",
-		"-m", "0777",
-		"-p",
+		"/bin/sh",
+		"/script/setup",
 	}
 	if err := p.createHelperPod(ActionTypeCreate, createCmdsForPath, name, path, node.Name); err != nil {
 		return nil, err
@@ -248,7 +247,7 @@ func (p *LocalPathProvisioner) Delete(pv *v1.PersistentVolume) (err error) {
 	}
 	if pv.Spec.PersistentVolumeReclaimPolicy != v1.PersistentVolumeReclaimRetain {
 		logrus.Infof("Deleting volume %v at %v:%v", pv.Name, node, path)
-		cleanupCmdsForPath := []string{"rm", "-rf"}
+		cleanupCmdsForPath := []string{"/bin/sh", "/script/teardown"}
 		if err := p.createHelperPod(ActionTypeDelete, cleanupCmdsForPath, pv.Name, path, node); err != nil {
 			logrus.Infof("clean up volume %v failed: %v", pv.Name, err)
 			return err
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -26,7 +26,8 @@ Kubernetes v1.12+.
 In this setup, the directory `/opt/local-path-provisioner` will be used across all the nodes as the path for provisioning (a.k.a, store the persistent volume data). The provisioner will be installed in `local-path-storage` namespace by default.
 
 ```
-kubectl apply -f https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml
+kuebctl apply -f deploy/local-path-storage.yaml
```
yasker, line 29: Hmm, we cannot do this. This will stop the one-line installation of Local Path Provisioner.

Better have the script in the config map by default.

provisioner.go
```diff
@@ -344,6 +344,11 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
 							ReadOnly:  false,
 							MountPath: "/data/",
 						},
+						{
+							Name:      "script",
+							ReadOnly:  false,
+							MountPath: "/tmp",
```
yasker, line 350: Don't use `/tmp`. It's a part of the helper pod anyway.

provisioner.go
```diff
@@ -358,6 +363,26 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
...
+				{
+					Name: "script",
+					VolumeSource: v1.VolumeSource{
+						ConfigMap: &v1.ConfigMapVolumeSource{
+							LocalObjectReference: v1.LocalObjectReference{
+								Name: "script",
+							},
+							Items: []v1.KeyToPath{
+								{
+									Key:  "add.sh",
```
yasker, line 375: We can call the script `setup`, `teardown`.

And it doesn't even need to be `.sh` since the interpreter should be specified at the first line of the script?

provisioner.go
```diff
@@ -323,7 +322,8 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
 	hostPathType := v1.HostPathDirectoryOrCreate
 	helperPod := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: string(action) + "-" + name,
+			Name:      string(action) + "-" + name,
+			Namespace: p.namespace,
```
yasker, line 326: This line should be a separate commit since normally we create the helper pod in the default namespace. Btw, why do we need this?

deploy/cm/add.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
yasker, line 1: how about just `#!/bin/bash`?

provisioner.go
nicktming: i have changed add.sh&del.sh to setup&teardown

provisioner.go
nicktming: i chaned /tmp to /script which is not a part of the helper pod.

README.md
nicktming: i removed cm dir and put the scripts into deploy/local-path-storage.yam with a script configmap by default.

deploy/cm/add.sh
nicktming: yes, i chanaged it to #!/bin/sh.

provisioner.go
nicktming: Since the helper pod needs to use the newly created configmap (script) which is in the same namespace as provisioner, then it is necessary to create/delete the helper pod with the provisioner's namespace.

yasker:
The code part looks good except for I think we can combine the configs.

A few other things:
1. Can you squash everything into one commit? Since it's one feature and is not complex enough to have functionally separated commits. You can use `git rebase -i` for that.
2. Can you update the documentation [Configuration section](https://github.com/rancher/local-path-provisioner#configuration), as well as [the example config](https://github.com/rancher/local-path-provisioner/blob/master/deploy/example-config.yaml)? Those can be in a separate commit or merged with the first one.
3. We need to update the [Helm Chart](https://github.com/rancher/local-path-provisioner/tree/master/deploy/chart). This should be included in the first commit.

deploy/local-path-storage.yaml
```diff
@@ -103,3 +103,19 @@ data:
                 ]
         }
 
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: script
```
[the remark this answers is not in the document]
nicktming: ok, it is done.

nicktming:
> The code part looks good except for I think we can combine the configs.
> 
> A few other things:
> 
>     1. Can you squash everything into one commit? Since it's one feature and is not complex enough to have functionally separated commits. You can use `git rebase -i` for that.
>  ok, it is done.

>     2. Can you update the documentation [Configuration section](https://github.com/rancher/local-path-provisioner#configuration), as well as [the example config](https://github.com/rancher/local-path-provisioner/blob/master/deploy/example-config.yaml)? Those can be in a separate commit or merged with the first one.
>  ok, it is done. i updated the documentation in readme and example config, i will commit if it needs more in the future.

yasker:
Hi @nicktming , how about the chart?

nicktming:
> Hi @nicktming , how about the chart?

ok, it is done.

yasker:
Merged, thanks @nicktming !

PR #105, opened 2020-07-09
OpenShift support - allow helper container to run in privileged mode for OCP

rakeshJn:
The local-path-provisioner does not work with Openshift, even after creating all possible security contexts. While debugging the issue, I found that the root cause is the helper container. It need to run in privileged mode so that it can create directories in the base path for each container needing PV.
This change helps with that. There is a new flag added to config.json called "privilegedHelper". By default it is set to false (for normal Kubernetes, it doesn't need privileged mode). For OCP, this must be set to true and with right security contexts, the provisioner works fine on OCP. Tested with OCP 4.3 and 4.4.
(please note, I can't do helm changes, if someone can contribute that, it will be great )

### DIFF
--- a/deploy/local-path-storage.yaml
+++ b/deploy/local-path-storage.yaml
@@ -100,7 +100,8 @@ data:
                         "node":"DEFAULT_PATH_FOR_NON_LISTED_NODES",
                         "paths":["/opt/local-path-provisioner"]
                 }
-                ]
+                ],
+                "privilegedHelper": false
         }
   setup: |-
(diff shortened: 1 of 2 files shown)


yasker:
Thanks for your PR @rakeshJn .

Since there is no need to dynamically change this parameter, I think it's better if we take it as a command-line parameter instead of a part of the config map.

adevjoe:
so we can add privileged in configmap to support openshift now.

like this
```
securityContext:
  privileged: true
```

PR #114, opened 2020-08-11
fix configmap name hardcode

### LINKED ISSUE #108: Custom teardown script doesn't work
I've tried to set a custom teardown script using Helm values:

```yaml
teardown: |-
  #!/bin/sh
  path=$1
  archived_path="$(dirname ${path})/archived-$(basename ${path})"
  mv ${path} ${archived_path}
```

Although the config map gets updated to the new teardown script, when I delete a pvc local-path-provisioner still deletes the pv folder instead of running the script.

Any help would be appreciated :-)

nicktming:
Fix configmap issue in v0.0.15. #108 #115

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,6 +35,7 @@ var (
 	DefaultKubeConfigFilePath = ".kube/config"
 	DefaultConfigFileKey      = "config.json"
 	DefaultConfigMapName      = "local-path-config"
+	FlagConfigMapName         = "configmap-name"
 )
 
 func cmdNotFound(c *cli.Context, command string) {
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -87,6 +88,11 @@ func StartCmd() cli.Command {
 				Usage: "Paths to a kubeconfig. Only required when it is out-of-cluster.",
 				Value: "",
 			},
+			cli.StringFlag{
+				Name:  FlagConfigMapName,
+				Usage: "Paths to a kubeconfig. Only required when it is out-of-cluster.",
```
yasker, line 93: Wrong usage.

nicktming: ok, it is done.

nicktming: and i think i need to update the helm chart and after that you can review.

nicktming: it works for me using helm deployment now.

Issue #84, opened 2020-05-11
local-path-provisioner doesn't work on clusters without internet


iirekm:
Fixing Helm chart is easy: add HELPER_IMAGE with env var to point to the proper address of busybox.

But additionally imagePullSecrets has to be added to the create-xxxx pods generated by the provisioner and there's no way to do it.

hwaastad:
@yasker this would be a minor change I think by extending the podpec in provisioner.go:
```
ImagePullSecrets: []v1.LocalObjectReference{
				{Name: p.pullSecret}, // should be loop of spec.imagepullsecrets
			},
```
We have met this issue our self having to use a private docker proxy.
I can add a PR for this one, but in my simple impl (since I have a limited knowledge of the go-client) I've added a pullproxy option to the provisioner (similar to the --helper-image). (Would probably be best if it was injected from the spec.imagepullsecrets from deployment,if possible)

yasker:
Seems we should add an image pull secret option in the cmdline.

vincent-herlemont:
If the PR https://github.com/rancher/local-path-provisioner/pull/119 is merged, this issue shouldn't be closed?

PR #119, opened 2020-08-18
add helper pull secret

hwaastad:
Trying to pass a pullsecret for the helper image.
Should help #84

### DIFF
--- a/deploy/chart/templates/deployment.yaml
+++ b/deploy/chart/templates/deployment.yaml
@@ -23,7 +23,11 @@ spec:
       serviceAccountName: {{ template "local-path-provisioner.serviceAccountName" . }}
       containers:
         - name: {{ .Chart.Name }}
+        {{- if .Values.privateRegistry.registryUrl }}
+          image: "{{ .Values.privateRegistry.registryUrl }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        {{- else }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        {{- end }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
             - local-path-provisioner
@@ -31,10 +35,16 @@ spec:
             - start
             - --config
             - /etc/config/config.json
+            - --service-account-name
+            - {{ template "local-path-provisioner.serviceAccountName" . }}
             - --provisioner-name
             - {{ template "local-path-provisioner.provisionerName" . }}
             - --helper-image
+          {{- if .Values.privateRegistry.registryUrl }}
+            - "{{ .Values.privateRegistry.registryUrl }}/{{ .Values.helperImage.repository }}:{{ .Values.helperImage.tag }}"
+          {{- else }}
             - "{{ .Values.helperImage.repository }}:{{ .Values.helperImage.tag }}"
+          {{- end }}
             - --configmap-name
             - {{ .Values.configmap.name }}
           volumeMounts:
--- a/main.go
+++ b/main.go
@@ -31,6 +31,8 @@ var (
 	FlagHelperImage        = "helper-image"
 	EnvHelperImage         = "HELPER_IMAGE"
 	DefaultHelperImage     = "busybox"
+	FlagServiceAccountName = "service-account-name"
+	EnvServiceAccountName  = "SERVICE_ACCOUNT_NAME"
 	FlagKubeconfig         = "kubeconfig"
 	DefaultConfigFileKey   = "config.json"
 	DefaultConfigMapName   = "local-path-config"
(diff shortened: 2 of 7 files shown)


will-beta:
Its helm chart needs some modifications.

main.go
```diff
@@ -93,6 +95,11 @@ func StartCmd() cli.Command {
 				Usage: "Required. Specify configmap name.",
 				Value: DefaultConfigMapName,
 			},
+			cli.StringFlag{
+				Name:   FlagPullSecret,
+				Usage:  "The pullsecret for helper images",
```
yasker, line 100: `ImagePullSecret for the helper image, if needed`

will-beta: Alernatively, we could write the `imagePullSecret` into the ServiceAccount in the chart, and use it in every pods, including the helper pods.

hwaastad: Good feedback. On the question on binding imagepullsecret to SA I feel this is an architecture design. In our setup this would work nicely. I´ll play around with the SA binding and either amend tmpl changes or new design (if sa binding is preferred)

yasker: Yeah, adding it to the service account is a good idea. Then we don't need to change the code of the provisioner. But we should document that or adding an option in the chart for that.

deploy/chart/templates/deployment.yaml
```diff
@@ -16,25 +16,35 @@ spec:
         app.kubernetes.io/name: {{ include "local-path-provisioner.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
-    {{- with .Values.imagePullSecrets }}
+    {{- if .Values.defaultSettings.registrySecret }}
```
will-beta, line 19: `.Values.defaultSettings.registrySecret` should only be used in the ServiceAccount. 
So changes from line 19 to 21 should be discarded.

will-beta (changes requested):
Nice work! Only one hunk that I think should be reverted before it gets merged.

deploy/chart/templates/deployment.yaml
hwaastad: Hi, I´ve reverted the change.

yasker:
LGTM. @will-beta ?

Also, this needs a rebase.

will-beta:
> LGTM. @will-beta ?
> 
> Also, this needs a rebase.

yasker:
Merged, thanks @hwaastad .

Can you add a few lines in the README.md so users know how to set the private registry?

hwaastad:
> Merged, thanks @hwaastad .
> 
> Can you add a few lines in the README.md so users know how to set the private registry?

Sure. A subsection under Configuration?

yasker:
@hwaastad Sure, thanks.

will-beta:
@hwaastad Could you help add support for existing imagePullSecrets? Thanks in advance.

hwaastad:
@will-beta, I can to a toyaml in the serviceaccount.yaml. (to add any defined imagepullsecrets in value.yaml`s). I`ll make another PR

deploy/chart/templates/deployment.yaml
```diff
@@ -23,18 +23,28 @@ spec:
...
+        {{- end }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
             - local-path-provisioner
             - --debug
             - start
             - --config
             - /etc/config/config.json
+            - --service-account-name
+            - {{ template "local-path-provisioner.serviceAccountName" . }}
```
kvaps, line 39: this is not working with current docker image: `rancher/local-path-provisioner:v0.0.17`:
```bash
# k logs local-path-provisioner-7ccbc4dfc9-sft56
Incorrect Usage: flag provided but not defined: -service-account-name
NAME:
   local-path-provisioner start - 
USAGE:
   local-path-provisioner start [command options] [arguments...]
OPTIONS:
   --config value            Required. Provisioner configuration file.
   --provisioner-name value  Required. Specify Provisioner name. (default: "rancher.io/local-path") [$PROVISIONER_NAME]
   --namespace value         Required. The namespace that Provisioner is running in (default: "local-path-storage") [$POD_NAMESPACE]
   --helper-image value      Required. The helper image used for create/delete directories on the host (default: "busybox") [$HELPER_IMAGE]
   --kubeconfig value        Paths to a kubeconfig. Only required when it is out-of-cluster.
   --configmap-name value    Required. Specify configmap name. (default: "local-path-config")
   
time="2020-09-09T12:31:07Z" level=fatal msg="Critical error: flag provided but not defined: -service-account-name" 
```

latest working commit for helm chart is `e4dfa34ae06b57667a854b216b5eba6f62f94cb7`

kvaps (commented):
-

deploy/chart/templates/deployment.yaml
hwaastad: Hi, currently only working on master. v0.0.17 does not support the sa option

yasker:
Hi @kvaps,

Once @hwaastad add the doc PR for the feature, I can cut a new release.

will-beta:
> Could you rebuild new docker image, and correct this problem?
> Thank you!

Also, could you make the chart searchable in the [helm hub](https://hub.helm.sh/)

PR #121, opened 2020-08-24
Fix kubeconfig priority

LeoLiuYan:


LeoLiuYan:
ping @yasker

.gitignore
```diff
@@ -4,3 +4,4 @@
 /dist
 *.swp
 .idea
+.vscode/
```
yasker, line 7: Can you separate this into another commit?

main.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	VERSION = "0.0.1"
+	VERSION = "0.0.17"
```
yasker, line 22: Don't need to change the VERSION variable manually since it should be changed while compiling the binary. You can run `./bin/local-path-provisioner --version` to verify.

main.go
```diff
@@ -110,17 +110,25 @@ func homeDir() string {
...
+			if _, err := os.Stat(f); err == nil {
+				return clientcmd.BuildConfigFromFlags("", f)
+			}
+		}
 	}
-	_, err := os.Stat(kubeconfig)
-	if err != nil {
-		return nil, err
+
+	if c, err := rest.InClusterConfig(); err == nil {
```
yasker, line 127: Hmm? So the priority would be:
1. ENV/RecommendedDir
2. In-cluster
3. RecommendedFileName 

Is this intended? I found it's weird that in-cluster is behind recommended dir but before the recommended file.

Also, do we still need `DefaultKubeConfigFilePath`?

LeoLiuYan: Oh... The priority of kubeconfig would be:

1.  Command args.  We could specify the --kubeconfig.
2. ENV.  We read from RecommendedConfigPathEnvVar.
3. In-cluster.
4. Recommended. We got by RecommendedHomeDir and RecommendedFileName.

yasker: That makes sense. Can you add this order to the commit message? Otherwise LGTM.

LeoLiuYan: OK.

PR #123, opened 2020-08-31
add default service account for debugging of out-of-cluster

nicktming:
there is no need to use service account for debugging of out-of-cluster, while it could not start under current condition, so adding a default service account for debugging of out-of-cluster.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ var (
 	EnvHelperImage         = "HELPER_IMAGE"
 	DefaultHelperImage     = "busybox"
 	FlagServiceAccountName = "service-account-name"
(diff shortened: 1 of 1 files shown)


yasker:
I think we should add the service account to the YAML file and chart instead, so it can be easily changed to another service account. By default, it should be `local-path-provisioner-service-account`.

nicktming:
> I think we should add the service account to the YAML file and chart instead, so it can be easily changed to another service account. By default, it should be `local-path-provisioner-service-account`.

ok, changed to `local-path-provisioner-service-account`.

yasker (approved):

PR #125, opened 2020-09-01
Adding block device support

tryggvil:
Adding support for Block VolumeMode in addition to FileSystem.

Block devices are created as files mounted on loopback.

Updating scripts to take in VolumeMode as first argument and path as second argument.

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -93,11 +93,33 @@ configmap:
   # specify the custom script for setup and teardown
   setup: |-
     #!/bin/sh
-    path=$1
-    mkdir -m 0777 -p ${path}
+    type=$1
+    path=$2
+    size=$3
+    if [ "${type}" == "Filesystem" ]; then
+      mkdir -m 0777 -p ${path}
+    elif [ "${type}" == "Block" ]; then
+      filepath="${path}-file"
+      LODEVICE=$(losetup -f)
+      truncate -s ${size} ${filepath}
+      chmod 666 ${filepath}
+      losetup ${LODEVICE} ${filepath}
+      ln -s  ${LODEVICE} ${path}
+    fi
 
   teardown: |-
     #!/bin/sh
-    path=$1
-    rm -rf ${path}
-
+    type=$1
+    path=$2
+    if [ "${type}" == "Filesystem" ]; then
+      rm -rf ${path}
+    elif [ "${type}" == "Block" ]; then
+      filepath="${path}-file"
+      // We need to use the first part of the string as losetup in busybox cuts off the full path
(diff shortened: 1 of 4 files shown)


deploy/chart/values.yaml
```diff
@@ -93,11 +93,33 @@ configmap:
   # specify the custom script for setup and teardown
   setup: |-
     #!/bin/sh
-    path=$1
-    mkdir -m 0777 -p ${path}
+    type=$1
```
yasker, line 96: We should use e.g. `--type` kind of parameter to separate them. Otherwise, it's easy to mix up the parameters.

In fact, we can just treat it as a boolean `--block`.

deploy/local-path-storage.yaml
```diff
@@ -66,6 +66,14 @@ spec:
         - start
         - --config
         - /etc/config/config.json
+        - --service-account-name
```
yasker, line 69: Those are default values. We don't need to add them explicitly? We can comment them out but leave them here, to let others know the existence of the options.

provisioner.go
```diff
@@ -356,6 +400,9 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
 						},
 					},
 					ImagePullPolicy: v1.PullIfNotPresent,
+					SecurityContext: &v1.SecurityContext{
```
yasker, line 403: I know it's due to `losetup` but I am concerned that for the user doesn't need this feature, they have to run another privileged container. It's better if we can find another way, e.g. expand the configmap. I have the same concern on https://github.com/rancher/local-path-provisioner/pull/126 too.

yasker (changes requested):
In general, I am OK with adding block volume support to the local path provisioner. But I have a few concerns:
1. The users that don't use this feature will still need to enable the privileged mode for the helper pod.
2. Should we support block volume created by losetup by default or should we provide an option to choose? I can imagining that we can support dmsetup as an alternative.
3. How many users going to use the local path provisioner to provision the raw block device? I think there might be some use cases but not sure if it's worth it to add it as a default option.

yasker:
@tryggvil Can you take a look at https://github.com/rancher/local-path-provisioner/pull/127 ? That approach looks good.

deploy/local-path-storage.yaml
tryggvil: In fact these are required when deploying straight from yaml, not the helm chart.
As you can see in https://github.com/rancher/local-path-provisioner/blob/master/main.go#L174 it errors out if flag is not set. So either handle empty flags with default values or pass in the flags by default,

innobead:
Closing the PR, it seems pending for a while. 

@tryggvil feel free to contribute again if #127 not enough for you. Thanks.

PR #126, opened 2020-09-02
enable xfs quota and add an xfs quota script example 

nicktming:
it is unsecure since pod could write unlimited size for current local-path-provisoner, so i made some changes to enable xfs quota for clusters which has xfs. And this is also compatiable with non-xfs clusters.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"strconv"
+
 	"github.com/Sirupsen/logrus"
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
@@ -193,11 +195,17 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	path := filepath.Join(basePath, folderName)
 	logrus.Infof("Creating volume %v at %v:%v", name, node.Name, path)
 
+	storage := pvc.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
+	iSizeInBytes, _ := storage.AsInt64()
+	sizeInBytes := strconv.FormatInt(iSizeInBytes, 10)
+
+	logrus.Infof("got sizeInBytes: %v, storage: %v, value: %v", sizeInBytes, storage.String(), storage.Value())
+
 	createCmdsForPath := []string{
 		"/bin/sh",
 		"/script/setup",
 	}
-	if err := p.createHelperPod(ActionTypeCreate, createCmdsForPath, name, path, node.Name); err != nil {
+	if err := p.createHelperPod(ActionTypeCreate, createCmdsForPath, name, path, node.Name, storage.Value()); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


provisioner.go
```diff
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"strconv"
```
yasker, line 13: It can be in the same section as `sync` and `time`.

provisioner.go
```diff
@@ -193,11 +195,17 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	path := filepath.Join(basePath, folderName)
 	logrus.Infof("Creating volume %v at %v:%v", name, node.Name, path)
 
+	storage := pvc.Spec.Resources.Requests[v1.ResourceName(v1.ResourceStorage)]
+	iSizeInBytes, _ := storage.AsInt64()
```
yasker, line 199: Better not ignore the error.

provisioner.go
```diff
@@ -193,11 +195,17 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
...
+	sizeInBytes := strconv.FormatInt(iSizeInBytes, 10)
+
+	logrus.Infof("got sizeInBytes: %v, storage: %v, value: %v", sizeInBytes, storage.String(), storage.Value())
```
yasker, line 202: Need to add volume name to the log.

provisioner.go
```diff
@@ -342,20 +352,34 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
...
-							MountPath: "/data/",
+							MountPath: parentDir,
 						},
 						{
 							Name:      "script",
 							ReadOnly:  false,
 							MountPath: "/script",
 						},
+						{
+							Name:      "xfs-quota-projects",
```
yasker, line 369: Hmm, those are too specific for XFS. Is there any way we can make it more generic?

provisioner.go
```diff
@@ -342,20 +352,34 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
 				{
 					Name:    "local-path-" + string(action),
 					Image:   p.helperImage,
-					Command: append(cmdsForPath, filepath.Join("/data/", volumeDir)),
+					Command: cmdsForPath,
+					Args:    []string{filepath.Join(parentDir, volumeDir), strconv.FormatInt(sizeInBytes, 10)},
```
yasker, line 356: It's better if we can use `--size` parameter instead of piling them up, otherwise it's easy to mix up the sequence.

provisioner.go
```diff
@@ -342,20 +352,34 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
...
+						},
+						{
+							Name:      "xfs-quota-projects",
+							SubPath:   "projid",
+							MountPath: "/etc/projid",
+						},
 					},
 					ImagePullPolicy: v1.PullIfNotPresent,
+					SecurityContext: &v1.SecurityContext{
+						Privileged: &enablePrivileged,
```
yasker, line 381: Because we need to access `/etc`? Seems unnecessary if the user is not running XFS.

provisioner.go
yasker: Also, this changes the input of the setup/teardown script format. We need to document the change.

yasker (changes requested):

innobead:
The need was resolved by #128 , closing.

PR #127, opened 2020-09-02
make helper pod configurable 

nicktming:
for #128

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,8 +19,7 @@ import (
 )
 
 var (
-	VERSION = "0.0.1"
-
+	VERSION                = "0.0.1"
 	FlagConfigFile         = "config"
 	FlagProvisionerName    = "provisioner-name"
 	EnvProvisionerName     = "PROVISIONER_NAME"
@@ -38,6 +37,8 @@ var (
 	DefaultConfigFileKey   = "config.json"
 	DefaultConfigMapName   = "local-path-config"
 	FlagConfigMapName      = "configmap-name"
+	FlagHelperPodFile      = "helper-pod-file"
+	DefaultHelperPodFile   = "helperPod.yaml"
 )
 
 func cmdNotFound(c *cli.Context, command string) {
@@ -101,6 +102,11 @@ func StartCmd() cli.Command {
 				EnvVar: EnvServiceAccountName,
 				Value:  DefaultServiceAccount,
 			},
+			cli.StringFlag{
+				Name:  FlagHelperPodFile,
+				Usage: "Paths to the Helper pod yaml file",
+				Value: "",
+			},
 		},
 		Action: func(c *cli.Context) {
 			if err := startDaemon(c); err != nil {
@@ -140,16 +146,16 @@ func loadConfig(kubeconfig string) (*rest.Config, error) {
 	return clientcmd.BuildConfigFromFlags("", kubeconfig)
 }
 
-func findConfigFileFromConfigMap(kubeClient clientset.Interface, namespace, configMapName string) (string, error) {
+func findConfigFileFromConfigMap(kubeClient clientset.Interface, namespace, configMapName, key string) (string, error) {
 	cm, err := kubeClient.CoreV1().ConfigMaps(namespace).Get(configMapName, metav1.GetOptions{})
 	if err != nil {
 		return "", err
 	}
-	configFile, ok := cm.Data[DefaultConfigFileKey]
+	value, ok := cm.Data[key]
 	if !ok {
-		return "", fmt.Errorf("%v is not exist in local-path-config ConfigMap", DefaultConfigFileKey)
+		return "", fmt.Errorf("%v is not exist in local-path-config ConfigMap", key)
 	}
-	return configFile, nil
+	return value, nil
 }
 
 func startDaemon(c *cli.Context) error {
@@ -185,7 +191,7 @@ func startDaemon(c *cli.Context) error {
 	}
 	configFile := c.String(FlagConfigFile)
 	if configFile == "" {
-		configFile, err = findConfigFileFromConfigMap(kubeClient, namespace, configMapName)
+		configFile, err = findConfigFileFromConfigMap(kubeClient, namespace, configMapName, DefaultConfigFileKey)
 		if err != nil {
 			return fmt.Errorf("invalid empty flag %v and it also does not exist at ConfigMap %v/%v with err: %v", FlagConfigFile, namespace, configMapName, err)
 		}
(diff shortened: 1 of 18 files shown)


mgoltzsche:
+1

I have similar requirements for [another project](https://github.com/mgoltzsche/jobcachefs) that are met by this PR.

main.go
```diff
@@ -99,6 +101,10 @@ func StartCmd() cli.Command {
 				Usage:  "Required. The ServiceAccountName for deployment",
 				EnvVar: EnvServiceAccountName,
 			},
+			cli.StringFlag{
+				Name:  FlagHelperPodFile,
+				Usage: "Required. The ServiceAccountName for deployment",
+			},
```
yasker, line 107: Is this required?

main.go
```diff
@@ -99,6 +101,10 @@ func StartCmd() cli.Command {
 				Usage:  "Required. The ServiceAccountName for deployment",
 				EnvVar: EnvServiceAccountName,
 			},
+			cli.StringFlag{
+				Name:  FlagHelperPodFile,
+				Usage: "Required. The ServiceAccountName for deployment",
```
yasker, line 106: Wrong usage.

provisioner.go
```diff
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"strconv"
```
yasker, line 13: Remove the empty line above.

provisioner.go
```diff
@@ -322,75 +335,66 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
...
+
+	helperPod.Namespace = p.namespace
+	helperPod.Spec.NodeName = node
+	helperPod.Spec.ServiceAccountName = p.serviceAccountName
+	helperPod.Spec.RestartPolicy = v1.RestartPolicyNever
+	helperPod.Spec.Tolerations = append(helperPod.Spec.Tolerations, tolerations...)
+	helperPod.Spec.Volumes = append(helperPod.Spec.Volumes, vols...)
+	helperPod.Spec.Containers[0].VolumeMounts = append(helperPod.Spec.Containers[0].VolumeMounts, volMnts...)
+	helperPod.Spec.Containers[0].Command = cmdsForPath
+	helperPod.Spec.Containers[0].Args = []string{filepath.Join(parentDir, volumeDir), strconv.FormatInt(sizeInBytes, 10), volumeMode}
```
yasker, line 397: We should add a prefix for each parameter now, it's getting confusing.

util.go
```diff
@@ -0,0 +1,37 @@
...
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
+)
+
+func loadFile(filepath string) (string, error) {
+	f, err := os.Open(filepath)
+	if err != nil {
+		return "", err
+	}
+	helperPodYaml, err := ioutil.ReadAll(f)
```
yasker, line 18: `defer` should be here.

yasker (changes requested):
Oh yeah, this approach is much better!

main.go
nicktming: i add a default value for this flag.

main.go
nicktming: ok, changed.

provisioner.go
nicktming: ok, done.

util.go
nicktming: ok, done.

provisioner.go
nicktming: i add a prefix for these parameters. @yasker could you check for this?

yasker: I meant, we can have the script accept parameters by name rather than just order, e.g.

`./setup --path <path> --size <size> --mode <mode>...`

Or we can just do 
`./setup -p <path> -s <size> -m <mode>...`

So the arguments are not solely decided by the order, make it easier to skip a few argument e.g. `size` if not needed, as well as make it clear to the caller how should it use the script.

yasker:
Also, use `rebase master` instead of `merge master` when using git.

provisioner.go
nicktming: ok, good idea, it is done now.

nicktming:
> Also, use `rebase master` instead of `merge master` when using git.

@yasker ok, done.

provisioner.go
```diff
@@ -322,75 +334,68 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
...
 				},
 			},
 		},
 	}
+	lpvVolumeMounts := []v1.VolumeMount{
+		{
+			Name:      "data",
+			ReadOnly:  false,
+			MountPath: parentDir,
+		},
```
mgoltzsche, line 374: A CLI option to specify the `MountPropagation` would also be great here.

mgoltzsche (changes requested):

provisioner.go
nicktming: @yasker what do you think?

mgoltzsche: Alternatively, following the template approach, the Pod template could specify the volume mount as well and the code here  could therefore only add the hardcoded volume mount when it is not specified.

mgoltzsche: I don't want to block this PR. So if you'd merge it as is I'll create another PR in which I'd also add a corresponding usage example to clarify the feature's purpose.

nicktming: > I don't want to block this PR. So if you'd merge it as is I'll create another PR in which I'd also add a corresponding usage example to clarify the feature's purpose.

i will add this flag soon, and  @yasker what do you think?

nicktming: > Alternatively, following the template approach, the Pod template could specify the volume mount as well which is then merged with/over the default, hard-coded volume mount above.
> 
> _It would help my use case already if you'd simply set `MountPropagation=v1.MountPropagationBidirectional` but I am not sure whether this could cause issues with other helpers and the default helper doesn't need it._

i add nodePathType and nodePathMountPropagation flag to enable this feature, you can check it. and @yasker what do you think?

main.go
```diff
@@ -21,23 +22,27 @@ import (
...
+	FlagServiceAccountName       = "service-account-name"
+	DefaultServiceAccount        = "local-path-provisioner-service-account"
+	EnvServiceAccountName        = "SERVICE_ACCOUNT_NAME"
+	FlagKubeconfig               = "kubeconfig"
+	DefaultConfigFileKey         = "config.json"
+	DefaultConfigMapName         = "local-path-config"
+	FlagConfigMapName            = "configmap-name"
+	FlagHelperPodFile            = "helper-pod-file"
+	DefaultHelperPodFile         = "helperPod.yaml"
+	FlagNodePathType             = "node-path-type"
```
yasker, line 44: Why do we need nodepathType?

main.go
```diff
@@ -101,6 +106,21 @@ func StartCmd() cli.Command {
...
+				Value: "",
+			},
+			cli.StringFlag{
+				Name:  FlagNodePathType,
+				Usage: "Paths to the Helper pod yaml file",
+				Value: string(v1.HostPathDirectoryOrCreate),
+			},
+			cli.StringFlag{
+				Name:  FlagNodePathMountPropagation,
+				Usage: "Paths to the Helper pod yaml file",
```
yasker, line 121: Wrong usage. And this should be another PR.

Also, I am not sure why we need mount propagation for the helper pod. @mgoltzsche can you explain your use case?

yasker (changes requested):
Sorry for the late review, was busy with Longhorn.

main.go
nicktming: since @mgoltzsche needs to open mount propagation flag, i made the node-path-type flag open to enable users to decide their own node path type such as BlockDevice, Socket etc.

main.go
mgoltzsche: @yasker my use case is [this](https://github.com/mgoltzsche/cache-provisioner): provision a cache as container image by running buildah within the helper pod to provision (and commit on PV deletion) the PV directory using overlayfs. However the overlayfs mount must be propagated from the helper pod to the host so that its contents can be mounted as volume to another pod after the helper pod terminated.

nicktming: @yasker @mgoltzsche  i removed mount propagation and node path type flag in this pr, after this pr is done, i will add these flags in next pr.

main.go
nicktming: i removed this flag in this pr.

main.go
mgoltzsche: @nicktming I'd do the next PR since other requirements came up and I could provide a corresponding usage example as well.

yasker (approved):

PR #133, opened 2020-09-12
WIP: Update docs

hwaastad:
Please comment.
Not sure about the docs since much of the changes are in helm deployment

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,44 @@ Check and follow the provisioner log using:
 $ kubectl -n local-path-storage logs -f -l app=local-path-provisioner
 ```
 
+### Air Gap Installation
+
+To be able to run the provisioner in an air gapped environment, you will need to provide a registry url, and a registry secret if needed.
+
+1. Create local-path-provisioner namespace
+
+```
+kubectl create namespace local-path-provisioner
+```
+2. If private registry needs authentication, create docker-registry secret in the local-path-provisioner namespace
+
+```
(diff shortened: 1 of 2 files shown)


will-beta (approved):

deploy/local-path-storage.yaml
```diff
@@ -58,12 +58,14 @@ spec:
       serviceAccountName: local-path-provisioner-service-account
       containers:
       - name: local-path-provisioner
-        image: rancher/local-path-provisioner:v0.0.17
+        image: rancher/local-path-provisioner:v0.0.18
```
innobead, line 61: Please rebase the PR to sync with the master code base

README.md
```diff
@@ -40,6 +40,44 @@ Check and follow the provisioner log using:
...
+```
+
+```
+command:
+        - local-path-provisioner
+        - --debug
+        - start
+        - --helper-image
+        - <REGISTRY_URL>/busybox
+        - --service-account-name
```
innobead, line 72: You don't need to specify service account name here, because the default value will be provided implictely

deploy/chart/local-path-storage-airgapped.yaml
```diff
@@ -0,0 +1,119 @@
+apiVersion: v1
```
innobead, line 1: It should be good enough to guide user how to modify the original deploy manifest instead of cloning another manfiest w/ less change for the template purpose.

innobead (changes requested):
Thanks for the contribution. Few coments to resolve.

PR #134, opened 2020-09-13
use multi-stage Dockerfile for packaging

tmclane:
Use a multi-stage dockerfile to build the image and avoid using the binary built on the host directly.

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,3 +1,9 @@
+FROM golang:1.15.2-alpine as builder
+
+COPY . /program
+WORKDIR /program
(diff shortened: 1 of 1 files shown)


yasker:
The binary is not built on the host directly. We use a tool called `dapper`(https://github.com/rancher/dapper) to build the image in the container. You can run `make` directly which will trigger `dapper` to build the provisioner for you.

tmclane:
Ah well it wasn't obvious to me when I looked.  I deployed it to my microk8s using the Dockerfile I submitted.

PR #135, opened 2020-09-15
Update helper image to point at Rancher-mirrored busybox

### LINKED ISSUE #120: Need to use rancher-mirrored busybox for helper image
The [default helper image](https://github.com/rancher/local-path-provisioner/blob/master/main.go#L33) pulls busybox from the docker library. This should be changed to pull from the Rancher image mirror (docker.io/rancher/busybox), in line with other project images.

brandond:
Fixes #120 - needed for k3s 1.19 release

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,7 +30,7 @@ var (
 	DefaultNamespace       = "local-path-storage"
 	FlagHelperImage        = "helper-image"
 	EnvHelperImage         = "HELPER_IMAGE"
-	DefaultHelperImage     = "busybox"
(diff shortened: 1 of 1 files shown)


brandond:
Note: I'll need a new release tagged with this so that I can use it in k3s.

cjellick:
@yasker you ok with this change?

brandond:
FYI we are going to use the HELPER_IMAGE override for now to avoid having to pull in all the recent changes. Would still like to have this in so that we can take advantage of it once we do update after the 1.19 release.

Issue #137, opened 2020-09-21
Unmounted storage protection


kvaps:
Hi the created volumes are having `type: DirectoryOrCreate` property.
But it might be unsafe in case if data storage is unmounted for some reason, because application can be started with the empty directory created in different place.

I'm sure that using `type: Directory` will solve this problem

https://github.com/rancher/local-path-provisioner/blob/b44b2c3b5d18772aa306d3fb1210b202c353ef09/provisioner.go#L205

yasker:
Sorry @kvaps I need to revert this change now. Somehow it doesn't work with my RKE cluster. I always get the error when trying to start the pod:
```
MountVolume.SetUp failed for volume "pvc-xxxx" : hostPath type check failed: /opt/local-path-provisioner/pvc-xxxxx is not a directory
```

kvaps:
But is it directory? :)

yasker:
Yes it is…

kvaps:
Weird, what kubelet vedsion do you have?

yasker:
I am using Rancher Kubernetes Engine (RKE), which is a containerized kubelet.

lllamnyp:
I'm out of context, but most likely if you're using RKE, you will need to specify somewhere in your cluster.yml something along the lines of
```
kubelet:
  extra_binds:
  - /opt/local-path-provisioner
```
refer to the RKE docs to get everything right.

kvaps:
Okay I prepared a new PR https://github.com/rancher/local-path-provisioner/pull/224

PR #138, opened 2020-09-21
Do not create directory if not found

kvaps:
Use `type: Directory` instead of `type: DirectoryOrCreate` allows to block running workload on provisioned directories, to avoid the situations when initial storage is unmounted or broken.

docker image containing the fix:
```
kvaps/local-path-provisioner:v0.0.17-fix-137
```

fixes https://github.com/rancher/local-path-provisioner/issues/137

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -202,7 +202,7 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	}
 
 	fs := v1.PersistentVolumeFilesystem
-	hostPathType := v1.HostPathDirectoryOrCreate
+	hostPathType := v1.HostPathDirectory
 	return &v1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,


kvaps:
Additionally I also added the check for filesystem type:

```yaml
configmap:

  setup: |
    #!/bin/sh
    path=$1
    dir=$(dirname "${path}")
    fslist='^(ext4|zfs|xfs)$'

    if [ -z "$(awk "\$2 == \"${dir}\" && \$3 ~ \"${fslist}\" {print \"ok\"; exit}" /proc/mounts)" ]; then
      echo "filesystem for "${dir}" is not match ${fslist}:" >&2
      awk "\$2 == \"${dir}\"" /proc/mounts >&2
      exit 1
    fi
    mkdir -m 0777 -p "${path}"

  teardown: |
    #!/bin/sh
    path=$1
    if [ ! -d "${path}" ]; then
      echo "directory "${path}" is not exists" >&2
      exit 1
    fi
    rm -rf "${path}"
```

yasker:
The PR itself looks good, but I am not sure why we want to check the fs type?

kvaps:
> I am not sure why we want to check the fs type?

This is exactly my case, because I have nodes booted via PXE with ephimeral overlayfs root.

See https://github.com/kvaps/kubefarm

There is a risk that volume might be created directly in the RAM partition and will be lost in the next reboot.

Filesystem type check guarantees that volume will be created on the normal "physical" filesystem.

yasker:
@kvaps Local path provisioner does offer the ability to set up customized locations for the persistent disk, so built on top of the ram disk should be preventable? Also, even if it's an ext4 filesystem, there is no guarantee it will be persistent since for some OS, certain directories will be cleaned up after reboot.

kvaps:
Setting the different location will not help if whole `/` (root filesystem) is ephimeral

yasker:
What's the expected behavior if the whole `/` is ephemeral? I don't get why checking fs type can help in this case. If there is a mixed deployment, trying to use RAM disk to create volume will result in volume creation failure and there won't be a retry on another node.

kvaps:
Just imagane you have:

```
/ - ephemeral overlayfs root (backed by tmpfs)
/opt/local-path-provisioner - is not mounted so located on the same ephemeral overlayfs root
```

let's assume the local-path-provisioner created the directory:

```
/opt/local-path-provisioner/pvc-xxxxx - still created on the same ephemeral overlayfs root
```

However pod is started some database and storing some data there.

During the restart we're having completely new root, `/opt/local-path-provisioner/pvc-xxxxx` - is not existing anymore, all data lost.

`type: DirectoryOrCreate` is creating new directory under the same path, pod is starting again with no errors. we're having new empty database.

- I want to avoid this case. By adding two checks:

- replace `type: DirectoryOrCreate` to `type: Directory` which will avoid to start pod if desired directory is not found (exactly this PR)
- add additional check for the provisioning new volumes, to avoid provisioning volumes on top of ephemeral overlayfs and tmpfs filesystems. (this is not so important if you're having normal root, but it is important if you're having ephemeral rootfs)

yasker:
`type: Directory` is a good change (don't understand why it doesn't though). But it's not going to solve the problem that you lost your data in this case.

In this case, that node should be excluded from the config map (https://github.com/rancher/local-path-provisioner#customize-the-configmap) to avoid losing data. Since even the check failed, the provisioner is not going to retry on a different node because the pod is already scheduled.

kvaps:
> But it's not going to solve the problem that you lost your data in this case.

At least it will return reilable error, instead of continue to work as if nothing had happened.

> In this case, that node should be excluded from the config map

I have all nodes like this

> even the check failed, the provisioner is not going to retry on a different node because the pod is already scheduled.

Sure, this is expected behavior. `WaitForFirstConsumer` allows to create PV on exact node where pod was scheduled.

My check allows to avoid situations when pod was scheduled to the wrong node by human error or if node has unmounted persistent storage for some reason.

This enhance is exact for my case, when nodes has ephimeral root. Unlike `type: Directory` it doesn't pretend to be merged into upstream.

PR #153, opened 2020-11-02
Fix error message related to the helper-pod-file flag

elek:
It's a very minor problem related to an error message introduced in #8740885, but I was confused.

When default configMap is used and `helperPod.yaml` is missing from the configMap, `main.go` complains about the `--config` flag instead of the `--helper-pod-file` flag.


yasker:
Thanks @elek for catching this.

I've rebased your PR to the `v0.0.19` branch which will be the base for the next release.

PR #161, opened 2020-12-06
use IfNotPresent image-pull-policy by default for helper-pod

unki:
I would like to suggest setting the _imagePullPolicy_ for the helper-pod to `IfNotPresent`.

Right now there is no image-tag given to the busybox-image, so `latest` is assumed - what implies _imagePullPolicy_ `Always` ([docs](https://kubernetes.io/docs/concepts/containers/images/#updating-images)). I've run quickly into the docker-hub-rate-limiting with that :)

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -137,6 +137,7 @@ configmap:
       containers:
       - name: helper-pod
         image: busybox
+        imagePullPolicy: IfNotPresent
 
 
 
--- a/deploy/example-config.yaml
+++ b/deploy/example-config.yaml
@@ -66,5 +66,6 @@ data:
           containers:
           - name: helper-pod
             image: busybox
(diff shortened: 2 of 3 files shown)


yasker:
Thanks @unki . I've made a branch for v0.0.20 and will merge your PR for it.

It will be released with v0.0.20 release later.

chattarajoy:
When can we expect this release to be available?

unki:
If you do not want to wait for the release - in the local-path-config configmap there is a key for _helperPod.yaml_, there you can set the `imagePullPolicy` to `IfNotPresent` as a workaround for the meantime until a new release is here.

My PR basically is just adapting the default-value when freshly deploying local-path-provisioner.

chattarajoy:
Thanks for the update, will try this out.

PR #166, opened 2020-12-30
Enhance helper Pod interface and configuration.

### LINKED ISSUE #164: Pass through PVC annotations to the helper Pod as env vars
To support advanced helper pod use cases:
* PersistentVolumeClaim annotation(s) that are passed through as env var(s) to the helper pod should be globally configurable.
* PVC name and namespace as well as PV name should always be provided as env vars to the helper Pod.

This would allow e.g. to configure a helper Pod that provisions PVs as caches based on a given cache key. The PVC/PV would only serve as session for a cache write which is applied to the given cache key on PV deletion. An example for such a helper is [cache-provisioner](https://github.com/mgoltzsche/cache-provisioner).

### LINKED ISSUE #165: Allow to configure helper Pod's data volume mount propagation
The `MountPropagation` of a helper Pod's `data` volume mount should be configurable.

Currently this volume mount is appended to the Pod template's volume mounts.
However the `data` volume mount should be merged with the one within the Pod template if it is provided.

My use case is a [cache provisioner](https://github.com/mgoltzsche/cache-provisioner) that provisions a PV by mounting a shared cache directory as overlayfs which requires `bidirectional` mount propagation of the `data` volume mount in order to let the PV mount outlive and exist outside of the Pod.

### LINKED ISSUE #175: Allow the helper pod to read params from env vars

### LINKED ISSUE #178: Make the helper Pod timeout configurable
...because depending on the helper Pod a different timeout may be needed.

My use case is [cache-provisioner](https://github.com/mgoltzsche/cache-provisioner) that requires the helper Pod to synchronize the volume contents with an image registry which may take more than 2 minutes.

mgoltzsche:
* Make the helper Pod receive only environment variables instead of args (this changes the interface in a non-backward compatible way but is simpler to use and potentially provides more backward compatibility in the future; #175).
* Adds the manager options `--pvc-annotation[-required]` to pass through annotations from the PVC to the PV and to the helper Pod (#164).
* Merge the helper Pod's `data` VolumeMount with the one provided with the template to be able to specify `mountPropagation` within the template (#165).
* Rename `helperPod.yaml` to `helper-pod.yaml` (more convenient and if we break sth we can break this as well).
* Expose `--helper-pod-timeout` option (#178).
* Provide a basic usage example of the new features (`examples/cache`).
* Support forceful termination of the manager binary (2xctrl+c - since this is annoying during development otherwise; #179).

Closes #164.
Closes #165.
Closes #175.
Closes #178.
Relates to #179.
Relates to #127.


provisioner.go
```diff
@@ -369,42 +440,48 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
...
+	helperPod.Spec.NodeName = o.Node
 	helperPod.Spec.ServiceAccountName = p.serviceAccountName
 	helperPod.Spec.RestartPolicy = v1.RestartPolicyNever
 	helperPod.Spec.Tolerations = append(helperPod.Spec.Tolerations, lpvTolerations...)
 	helperPod.Spec.Volumes = append(helperPod.Spec.Volumes, lpvVolumes...)
-	helperPod.Spec.Containers[0].VolumeMounts = append(helperPod.Spec.Containers[0].VolumeMounts, lpvVolumeMounts...)
-	helperPod.Spec.Containers[0].Command = cmdsForPath
-	helperPod.Spec.Containers[0].Args = []string{"-p", filepath.Join(parentDir, volumeDir),
-		"-s", strconv.FormatInt(sizeInBytes, 10),
-		"-m", volumeMode}
```
mgoltzsche, line 407: This is a breaking change!
Reading these options within a setup/teardown script is pretty cumbersome as the previous examples show.
Therefore I propose to use environment variables for these parameters. Also env vars potentially provide better backward compatibility than options (imagine sb builds a binary using a CLI framework: it would fail by default when new/unknown options are added while, when using env vars, this is not a problem).
However we could leave the options here for now in addition to the env vars, deprecate the options somehow and remove them later. Though eventually we can remove them now since the feature is pretty young anyway and not many people may rely on it / not many systems have to be changed.
WDYT?

@nicktming this affects the changes you introduced in #127.

main.go
```diff
@@ -19,26 +21,34 @@ import (
...
+	FlagServiceAccountName    = "service-account-name"
+	DefaultServiceAccount     = "local-path-provisioner-service-account"
+	EnvServiceAccountName     = "SERVICE_ACCOUNT_NAME"
+	FlagKubeconfig            = "kubeconfig"
+	DefaultConfigFileKey      = "config.json"
+	DefaultConfigMapName      = "local-path-config"
+	FlagConfigMapName         = "configmap-name"
+	EnvConfigMapName          = "CONFIGMAP_NAME"
+	FlagHelperPodFile         = "helper-pod-file"
+	DefaultHelperPodFile      = "helper-pod.yaml"
```
mgoltzsche, line 44: This is another breaking change. I could revert that one but the lower case notation is more convenient and if we break sth here already we might as well break this one now.

mgoltzsche:
@yasker please have a look

mgoltzsche:
I just rebased this PR and resolved the conflicts.

innobead:
Thanks for the contribution. 

It looks you want to solve several issues or enhancements, so I would suggest to seperate this PR first, so we can focus on each problem you are resolving and it will be better for review.

mgoltzsche:
@innobead okay, I moved the `mountPropagation` feature into a new PR now: #176 - though I've also added the env var support to it (to reduce merge conflicts within other PRs) but in a backward compatible way.

LarsBingBong:
In regards to https://github.com/rancher/local-path-provisioner/pull/177 having been merged. I'm getting the following err:

image

In the local-path helper `Pod`. I deployed from https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml as the ReadMe describes - and I did so today.

Sure enough I haven't provided this `VOL_DIR` parameter anywhere. Where am I to do so? I can't seem to find it described anywhere.

Thank you very much.

LarsBingBong:
Yeah! That made it work! Thank you very much @derekbit

PR #167, opened 2021-01-01
Disable (un-needed) leader election to reduce pod restarts

mnorrsken:
This is a very quick fix to disable leader election that is not needed for non-replicated storage classes. This stops pod restarts because of re-elections.


innobead:
LGTM, and for the current design, we don't need to have multiple pods to handle the provision work.

innobead (approved):

PR #176, opened 2021-03-06
Allow mountPropagation configuration by merging the helper Pod template's volumeMount

### LINKED ISSUE #165: Allow to configure helper Pod's data volume mount propagation

### LINKED ISSUE #175: Allow the helper pod to read params from env vars

mgoltzsche:
The helper Pod template's `data` volumeMount is merged with the defaults in order to allow users to specify the `mountPropagation` within the template.

Split from PR #166.
Closes #165.
Relates to #127.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -31,6 +31,10 @@ const (
 	KeyNode = "kubernetes.io/hostname"
 
 	NodeDefaultNonListedNodes = "DEFAULT_PATH_FOR_NON_LISTED_NODES"
+
+	helperScriptDir     = "/script"
+	helperDataVolName   = "data"
+	helperScriptVolName = "script"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


mgoltzsche:
hm, the e2e test `TestPVCTestSuite/TestPodWithSecurityContext` fails.
For some reason the `local-path-test` Pod has no permissions to write into the volume and logs:
```
uid=1000(jenkins) gid=1000(jenkins) groups=1000(jenkins)
total 8
drwxr-xr-x 2 root root 4096 Mar  6 22:18 .
drwxr-xr-x 1 root root 4096 Mar  6 22:18 ..
Hello from local-path-test
cp: cannot create regular file '/test/test.txt': Permission denied
```

mgoltzsche:
Fixed. This is ready for review now.

pzeinlinger:
Can we get this change merged anytime soon?

derekbit:
> Can we get this change merged anytime soon?

Thanks for the notification. I will review it.

pzeinlinger:
Hi @derekbit 
have you had a chance to review this PR yet?

provisioner.go
```diff
@@ -369,42 +380,40 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmdsForPath []
...
-	}
 	lpvTolerations := []v1.Toleration{
 		{
 			Operator: v1.TolerationOpExists,
 		},
 	}
 	helperPod := p.helperPod.DeepCopy()
 
+	scriptMount := addVolumeMount(&helperPod.Spec.Containers[0].VolumeMounts, helperScriptVolName, helperScriptDir)
+	scriptMount.MountPath = helperScriptDir
```
derekbit, line 391: Why is scriptMount.MountPath assigned again?

mgoltzsche: `addVolumeMount` would not overwrite the `MountPath` that the Pod template specifies (there are potential use cases where the volume path within the container must equal the path on the host and therefore users would want to change the path as well within the Pod template).

By explicitly assigning `MountPath` of the script volume mount I ensured that users cannot overwrite the `/script` path within the Pod template in the same way since the `cmd`` that the calling code provides relies on it (I didn't see a reason why users would want to configure a script path other than the hardcoded one).

derekbit: Make sense. Thank you.

derekbit (approved):
LGTM

PR #177, opened 2021-03-06
Provide helper Pod params as env vars.

### LINKED ISSUE #175: Allow the helper pod to read params from env vars

mgoltzsche:
For backward compatibility the CLI options are still provided.

Closes #175
Split from PR #166


mgoltzsche:
@derekbit this PR is also ready for review - I just rebased it.

derekbit:
@mgoltzsche 
Could you elaborate more the purpose of moving the the variables (absolutePath, sizeInBytes and volMode) to env variables? This will be helpful for understanding the series of the PRs clearly.

mgoltzsche:
It reduces the amount of redundant code needed to configure a helper pod.

derekbit (approved):
LGTM

PR #188, opened 2021-05-01
[Feat] Add  the new feature to support the air-gapped situation

Aisuko:
Signed-off-by: aisuko <urakiny@gmail.com>

Under the offline situation, we can only use the images that were pre-load to the host. The specification of `helperPod.yaml` used the default value `imagePullPolicy` which equals `Always` will let the help pod failed to work. So, we should declare the specification of the image pull policy as `IfNotPresent`.

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -137,6 +137,7 @@ configmap:
       containers:
       - name: helper-pod
         image: busybox
+        imagePullPolicy: IfNotPresent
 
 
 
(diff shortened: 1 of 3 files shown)


Aisuko:
Please confirm bump the chart version or not. I saw the chart version follow the application version. So, I did not update the version of the chart

Aisuko:
@innobead Thanks

tallaxes:
The reason `imagePullPolicy` had defaulted to `Always` is because of missing image tag. It is fine to set it explicitly, but it would be even better to include specific image tag, e.g. `busybox:1.33.1` ...

Aisuko:
@tallaxes Welcome any PR to implement this

PR #189, opened 2021-05-11
indicate volume expansion is possible

ekristen:
By adding `allowVolumeExpansion` this will prevent errors if someone changes the PVC capacity size on an already provisioned PVC, since local-path-provisioner does not enforce disk usage anyways, it seems reasonable to allow the value to change without throwing an error.

However this will not update the status of the PVC, it'll still show the original amount. The project would need some sort of resource watch and update status to reflect status from spec.

### DIFF
--- a/debug/config.yaml
+++ b/debug/config.yaml
@@ -11,6 +11,7 @@ metadata:
 provisioner: rancher.io/local-path
 volumeBindingMode: WaitForFirstConsumer
 reclaimPolicy: Delete
+allowVolumeExpansion: true
 
 ---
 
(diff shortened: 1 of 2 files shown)


ekristen:
Issue opened and referenced. My point if I haven't made it clear is that since the local path provisioner doesn't enforce disk usage, or check for disk usage, or provisions actual disk space, why prevent the volume size for changing?

In my use case I am programmatically creating PVCs using the storage class and if the underlying setting in my app changes for the default disk size all the api calls break because the provisioner says it doesn't support expansion when in fact it could and should since it doesn't actually do any provisioning around size anyways.  :) 

Thanks for your time.

flokli:
What's the status of this?

If we don't enforce resource limits/quotas anyways, could we just set this, as this PR already does?

innobead (approved):
LGTM

innobead:
> Issue opened and referenced. My point if I haven't made it clear is that since the local path provisioner doesn't enforce disk usage, or check for disk usage, or provisions actual disk space, why prevent the volume size for changing?
> 
> In my use case I am programmatically creating PVCs using the storage class and if the underlying setting in my app changes for the default disk size all the api calls break because the provisioner says it doesn't support expansion when in fact it could and should since it doesn't actually do any provisioning around size anyways. :)
> 
> Thanks for your time.

@ekristen thanks for the explanation and contribution :+1:

PR #196, opened 2021-07-12
Use busybox:1.33.1 in helper-pod

tallaxes:
Use busybox:1.33.1 - which is currently the latest available. The use of specific image tag also changes `imagePullPolicy` default from `Always` to `IfNotPresent` - which, among other things, behaves better in aigrap, see https://github.com/rancher/local-path-provisioner/pull/188#issuecomment-877459415.

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -136,10 +136,4 @@ configmap:
     spec:
       containers:
       - name: helper-pod
-        image: busybox
-        imagePullPolicy: IfNotPresent
-
-
-
-
-
+        image: busybox:1.33.1
--- a/deploy/example-config.yaml
+++ b/deploy/example-config.yaml
@@ -65,7 +65,4 @@ data:
     spec:
       containers:
       - name: helper-pod
-        image: busybox
-        imagePullPolicy: IfNotPresent
-
-
+        image: busybox:1.33.1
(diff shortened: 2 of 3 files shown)


deploy/chart/values.yaml
```diff
@@ -136,10 +136,4 @@ configmap:
...
       containers:
       - name: helper-pod
-        image: busybox
-        imagePullPolicy: IfNotPresent
-
-
-
-
-
+        image: busybox:1.33.1
```
derekbit, line 139: Looks to me. But suggest to keep `imagePullPolicy: IfNotPresent` explicitly.

deploy/example-config.yaml
```diff
@@ -65,7 +65,4 @@ data:
     spec:
       containers:
       - name: helper-pod
-        image: busybox
-        imagePullPolicy: IfNotPresent
-
-
+        image: busybox:1.33.1
```
derekbit, line 68: Same as the above.

deploy/local-path-storage.yaml
```diff
@@ -152,7 +152,4 @@ data:
     spec:
       containers:
       - name: helper-pod
-        image: busybox
-        imagePullPolicy: IfNotPresent
-
-
+        image: busybox:1.33.1
```
derekbit, line 155: Same as the above.

derekbit:
In current design, the latest busybox image is pulled. For security, I think it's fair.
User can use a specific version and imagePullPolicy for their use cases.

PR #213, opened 2021-10-11
feat: Make command timeout configurable

jaysonsantos:
Hey there, I'd like to introduce this change in an effort to circunvene agents that have [this problem](https://github.com/containerd/containerd/issues/4604) (in my case probably slow disk) that make the containers take longer time to spin up and the hardcoded 120 seconds make a few pvcs not being provisioned when the helper container takes too much time to start.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -37,9 +37,11 @@ const (
 	helperScriptVolName = "script"
 )
 
-var (
-	CmdTimeoutCounts = 120
+const (
+	defaultCmdTimeoutSeconds = 120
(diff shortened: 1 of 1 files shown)


derekbit:
@jaysonsantos 

Thanks for your contribution.
Are you willing to update and resubmit the PR?
The configurable timeout is helpful for some use cases.

jaysonsantos:
@derekbit just rebased it

derekbit:
In general, LGTM.

PR #217, opened 2021-11-11
Add s390x arch support

rohitsakala:
Add s390x support to rancher/rancher

Part of https://github.com/rancher/rancher/issues/35501


derekbit:
LGTM.

Issue #214, opened 2021-10-14
pvc provision does not create more then 4 helper PODs at the time 


talcoh2x:
Hi, maybe you can help me.
I'm using with local-path-provisioner and I really love it :)
my main problem is that once I'm creating 16/20 differents PVC  ( one pvc for each worker  ) I can see that in MAX i have 4 helper PODs . means that I block on 4 POD's creation each time.
there is a limit or something like that in k8s?
what can be the reason for that ?

skyoo2003:
@talcoh2x 

The PR https://github.com/rancher/local-path-provisioner/pull/222 added a flag like '--worker-threads' to change provisioning concurrency. so, you can test it after the feature is released.

derekbit:
Already merged #222.

PR #222, opened 2021-12-16
Added options for worker threads and provisioning/deletion retry counts

skyoo2003:
The 'sig-storage-lib-external-provisioner' provides an option for volume provisioning and deletion concurrency. so, I added an option like '--worker-threads' to allow users to control provisioning concurrency. Failed provision/delete threshold values are zero, so the provisioner can complete provisioning and deletion.
Please review.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,28 +19,32 @@ import (
 )
 
 var (
-	VERSION                = "0.0.1"
-	FlagConfigFile         = "config"
-	FlagProvisionerName    = "provisioner-name"
-	EnvProvisionerName     = "PROVISIONER_NAME"
-	DefaultProvisionerName = "rancher.io/local-path"
-	FlagNamespace          = "namespace"
-	EnvNamespace           = "POD_NAMESPACE"
-	DefaultNamespace       = "local-path-storage"
-	FlagHelperImage        = "helper-image"
-	EnvHelperImage         = "HELPER_IMAGE"
-	DefaultHelperImage     = "rancher/library-busybox:1.32.1"
-	FlagServiceAccountName = "service-account-name"
-	DefaultServiceAccount  = "local-path-provisioner-service-account"
-	EnvServiceAccountName  = "SERVICE_ACCOUNT_NAME"
-	FlagKubeconfig         = "kubeconfig"
-	DefaultConfigFileKey   = "config.json"
-	DefaultConfigMapName   = "local-path-config"
-	FlagConfigMapName      = "configmap-name"
-	FlagHelperPodFile      = "helper-pod-file"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -232,6 +239,9 @@ func startDaemon(c *cli.Context) error {
 		provisioner,
 		serverVersion.GitVersion,
 		pvController.LeaderElection(false),
+		pvController.FailedProvisionThreshold(0),
+		pvController.FailedDeleteThreshold(0),
```
FranciscoMSM, line 243: The default values for these should be 15. 0 means infinite retrying.

Consider adding a command-line argument to control these defaults as well.

skyoo2003: @FranciscoMSM Thanks for your feedback.

The number of provisioning and deletion retry thresholds has also been changed to be controlled by user options like (--provisioning-retry-count, --deletion-retry-count).

skyoo2003:
I think it's probably related to the issue https://github.com/rancher/local-path-provisioner/issues/214.

FranciscoMSM (commented):
lgtm

I'm not a maintainer though :)

derekbit:
LGTM.
Thanks for your contribution.

skyoo2003:
@derekbit

Thank you for the positive feedback. 😀
Could you let me know when this PR can be mergred and released?

main.go
```diff
@@ -232,6 +253,9 @@ func startDaemon(c *cli.Context) error {
 		provisioner,
 		serverVersion.GitVersion,
 		pvController.LeaderElection(false),
+		pvController.FailedProvisionThreshold(c.Int(FlagProvisioningRetryCount)),
+		pvController.FailedDeleteThreshold(c.Int(FlagDeletionRetryCount)),
+		pvController.Threadiness(c.Int(FlagWorkerThreads)),
```
derekbit, line 258: Should we check if they are non-negative integers before creating the `NewProvisionController`?

derekbit:
@skyoo2003 
I will merge it after resolving the minor issue and plan to release in mid January next year.

main.go
skyoo2003: I agree. I'm gonna change this.

derekbit:
LGTM. Thank you. :)

talcoh2x:
Cool!! That what I search for. I didn't find how to do it.
I'll test it in few days on my Env . Can you please elaborate about the minor issue? @derekbit

derekbit:
@talcoh2x 
The minor issue I mentioned is the illegal value checks and fixed by @skyoo2003.
This PR is merged and will be in the next release.

talcoh2x:
@derekbit  thanks!

PR #224, opened 2021-12-22
Do not create directory if not found

kvaps:
Same as https://github.com/rancher/local-path-provisioner/pull/138, second attempt

Use `type: Directory` instead of `type: DirectoryOrCreate` allows to block running workload on provisioned directories, to avoid the situations when initial storage is unmounted or broken.

fixes https://github.com/rancher/local-path-provisioner/issues/137

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -215,7 +215,7 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	}
 
 	fs := v1.PersistentVolumeFilesystem
-	hostPathType := v1.HostPathDirectoryOrCreate
+	hostPathType := v1.HostPathDirectory
 
 	valueNode, ok := node.GetLabels()[KeyNode]
 	if !ok {


david-suba:
What is status of this PR, please? I don't think that provisiner should create directory  if it doesn't exist before. At least option to configure this behaviour would be nice.

derekbit:
@kvaps @david-suba 
Agree with your idea. If the parent directory does not exist, the volume should not be provisioned.

But I think we should change the `hostPathType` to `v1.HostPathDirectory` in `createHelperPod()` ([code](https://github.com/rancher/local-path-provisioner/blob/master/provisioner.go#LL458C2-L458C46)) rather than in the `Provision()` ([code](https://github.com/rancher/local-path-provisioner/blob/master/provisioner.go#L292)).

The change will change the previous behavior. I would suggest the `hostPathType` can be gated by a PVC annotation or other methods. WDYT?

derekbit:
v0.0.24 will be released by this Wed. Put in v0.0.25 if can't catch up with v0.0.24.

PR #187, opened 2021-04-13
Use the node label for NodeAffinity rule

uempfel:
closes #186

PR #227, opened 2022-01-31
Use the node label for NodeAffinity rule for helper pod

Sudokamikaze:
Backport feature: https://github.com/rancher/local-path-provisioner/pull/187 for helper pod

We experienced issue when helper pod assigns on node based on hostname, instead of actual nodeName on EKS


Sudokamikaze:
@yasker
@innobead

Could you please review it when you have some time? Thanks

PR #228, opened 2022-02-02
feat: add resource limits

githubcdr:
- Limit resources for deployment


deploy/local-path-storage.yaml
```diff
@@ -71,8 +71,11 @@ spec:
             - --config
             - /etc/config/config.json
           resources:
+            requests:
+              cpu: 50m
+              memory: 128Mi
             limits:
-              cpu: 1
+              cpu: 250m
```
derekbit, line 78: Why choose 250m as the limit?

PR #252, opened 2022-06-16
drop unnecessary imagePullSecrets setting on charts for deplyment

tgfree7:
Setting imagePullSecrets and serviceAccountName in the same time, is unnecessary, and may set the wrong secrets which will disable the sa.


justusbunsi (approved):
LGTM from an external point of view 😃

justusbunsi:
Bad stale bot, bad. 😁

derekbit:
> Bad stale bot, bad. 😁

It's annoying. Feel free to reopen it.

PR #260, opened 2022-07-29
Add option to create local volumes instead of hostPath

### LINKED ISSUE #85: enable velero backups by using local instead of hostpath

AnthonyEnr1quez:
resolve #85

Took the changes from #91 and added them to be enabled by ~~a config value~~ an annotation on the pvc. Default behavior is still to use hostPath volumes. By using an annotation, we can allow the user to create both `hostPath` and `local` volumes, instead of just locking down to one.

Made some small changes to the README, but will update the README and examples further if code changes are ok from maintainers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Local Path Provisioner provides a way for the Kubernetes users to utilize the lo
 ## Compare to built-in Local Persistent Volume feature in Kubernetes
 
 ### Pros
-Dynamic provisioning the volume using [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) or [local volume](https://kubernetes.io/docs/concepts/storage/volumes/#local).
+Dynamic provisioning the volume using [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) or [local](https://kubernetes.io/docs/concepts/storage/volumes/#local).
 * Currently the Kubernetes [Local Volume provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner) cannot do dynamic provisioning for the local volumes.
 
 ### Cons
--- a/README.md
+++ b/README.md
@@ -3,12 +3,12 @@
 
 ## Overview
 
-Local Path Provisioner provides a way for the Kubernetes users to utilize the local storage in each node. Based on the user configuration, the Local Path Provisioner will create `hostPath` based persistent volume on the node automatically. It utilizes the features introduced by Kubernetes [Local Persistent Volume feature](https://kubernetes.io/blog/2018/04/13/local-persistent-volumes-beta/), but makes it a simpler solution than the built-in `local` volume feature in Kubernetes.
+Local Path Provisioner provides a way for the Kubernetes users to utilize the local storage in each node. Based on the user configuration, the Local Path Provisioner will create either `hostPath` or `local` based persistent volume on the node automatically. It utilizes the features introduced by Kubernetes [Local Persistent Volume feature](https://kubernetes.io/blog/2018/04/13/local-persistent-volumes-beta/), but makes it a simpler solution than the built-in `local` volume feature in Kubernetes.
 
 ## Compare to built-in Local Persistent Volume feature in Kubernetes
 
 ### Pros
-Dynamic provisioning the volume using hostPath.
+Dynamic provisioning the volume using [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) or [local volume](https://kubernetes.io/docs/concepts/storage/volumes/#local).
 * Currently the Kubernetes [Local Volume provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner) cannot do dynamic provisioning for the local volumes.
 
 ### Cons
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Local Path Provisioner provides a way for the Kubernetes users to utilize the lo
 ### Pros
 Dynamic provisioning the volume using [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) or [local](https://kubernetes.io/docs/concepts/storage/volumes/#local).
 * Currently the Kubernetes [Local Volume provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner) cannot do dynamic provisioning for the local volumes.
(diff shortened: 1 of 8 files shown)


derekbit:
@AnthonyEnr1quez Looks good. Can you solve the conflict, thank you.

AnthonyEnr1quez:
Changed the status to Draft, will fix conflicts and add more docs/examples and tests

AnthonyEnr1quez:
@derekbit Fixed the conflicts and added examples and an e2e test. Changed the mechanism to use local volume to annotations on the pvc.

examples/pvc-with-local-volume/pvc.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: local-volume-pvc
+  annotations:
+    volumeType: local
```
AnthonyEnr1quez, line 6: We could rename this variable, do we want something like `rancher.io/local-path-provisioner/volume-type`

AnthonyEnr1quez:
If these changes are ok, I might refactor the e2e tests to be a bit cleaner and remove the dependency on testify suite

examples/pvc-with-local-volume/pvc.yaml
derekbit: I think this is fine. Any concern?

AnthonyEnr1quez: Not a big concern, just thinking that if we lock down the annotation key to use a rancher prefix, it is 100% tied to this provisioner.

Could be some weird edge case where a person uses `volumeType` annotation as some sort of internal value for tracking, not knowing that it changes pv creation behavior.

Marcel-Lambacher:
@AnthonyEnr1quez Do you need support here?
I already added support for PVC annotations into onechart - https://github.com/gimlet-io/onechart/pull/51

I'm really looking forward for this feature, since it's blocking my from creating an end-to-end backup strategy.
But it's always easy to say: "what's the status", instead maybe I can support you here?

maximeborges:
As @Marcel-Lambacher, also waiting for this feature for better handling of backups.
I can also provide help if needed, any status @AnthonyEnr1quez?

AnthonyEnr1quez:
@Marcel-Lambacher @maximeborges we should be good, just waiting for a response to https://github.com/rancher/local-path-provisioner/pull/260#discussion_r961817716

examples/pvc-with-local-volume/pvc.yaml
AnthonyEnr1quez: @derekbit If we are fine leaving this as-is, I can close this comment and merge?

README.md
```diff
@@ -3,12 +3,12 @@
 
 ## Overview
 
-Local Path Provisioner provides a way for the Kubernetes users to utilize the local storage in each node. Based on the user configuration, the Local Path Provisioner will create `hostPath` based persistent volume on the node automatically. It utilizes the features introduced by Kubernetes [Local Persistent Volume feature](https://kubernetes.io/blog/2018/04/13/local-persistent-volumes-beta/), but makes it a simpler solution than the built-in `local` volume feature in Kubernetes.
+Local Path Provisioner provides a way for the Kubernetes users to utilize the local storage in each node. Based on the user configuration, the Local Path Provisioner will create either `hostPath` or `local` based persistent volume on the node automatically. It utilizes the features introduced by Kubernetes [Local Persistent Volume feature](https://kubernetes.io/blog/2018/04/13/local-persistent-volumes-beta/), but makes it a simpler solution than the built-in `local` volume feature in Kubernetes.
```
derekbit, line 6: I'm thinking if we can mark `local` as an experimental feature and see any issues from other users?

derekbit (commented):
In general, LGTM.

provisioner.go
```diff
@@ -229,7 +228,23 @@ func (p *LocalPathProvisioner) Provision(opts pvController.ProvisionOptions) (*v
 	}
 
 	fs := v1.PersistentVolumeFilesystem
-	hostPathType := v1.HostPathDirectoryOrCreate
+
+	var pvs v1.PersistentVolumeSource
+	if val, ok := opts.PVC.GetAnnotations()["volumeType"]; ok && strings.ToLower(val) == "local" {
```
derekbit, line 233: This is good. The existing volumes still use `hostPath`. Thank you.

README.md
AnthonyEnr1quez: Added https://github.com/rancher/local-path-provisioner/pull/260/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R13

derekbit:

fabioandreola:
Hi, do we have a date when this will be released?

derekbit:
We will release the new version by the end of this week. I'm testing and fixing some issues.

derekbit:
v0.0.23 is released. Thank you and any feedback is appreciated.

AnthonyEnr1quez:
@Marcel-Lambacher @maximeborges @fabioandreola if you need help converting volumes, see https://github.com/rancher/local-path-provisioner/issues/85#issuecomment-1292578697

fabioandreola:
> @Marcel-Lambacher @maximeborges @fabioandreola if you need help converting volumes, see [#85 (comment)](https://github.com/rancher/local-path-provisioner/issues/85#issuecomment-1292578697)

Thanks. I want to report that with this change Velero PVC backups are working as expected

PR #264, opened 2022-09-16
Fix: hostname label != node name results in cleanup HelperPod not being scheduled

rcasaleiro:
In situations where the hostname has been overriden through the kubelets **--hostname-override** flag or when the node name is a FQDN and kubelet takes the default hostname of the node (non-FQDN).
Although creation of PV's will work just fine, the cleanup of said PV fails as the value for the node name given to the Delete HelperPod is taken from the hostname label affinity.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -349,19 +349,19 @@ func (p *LocalPathProvisioner) Delete(pv *v1.PersistentVolume) (err error) {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


derekbit:
@rcasaleiro Can you help resolve the conflict?

jan-g:
See #413 and attached PR.

Issue #274, opened 2022-10-30
[Bug] [Helm] sharedFilesystemPath Missing


meln5674:
The config.json field sharedFilesystemPath is [not set in the ConfigMap template](https://github.com/rancher/local-path-provisioner/blob/master/deploy/chart/local-path-provisioner/templates/configmap.yaml#L10), despite [being present in the values.yaml](https://github.com/rancher/local-path-provisioner/blob/master/deploy/chart/local-path-provisioner/values.yaml#L67). This means that instances deployed by helm cannot use the ReadWriteMany access mode.

derekbit:
Thank you for the fix. I think we should generate the deply manifest from the heml chart to avoid this kind of bug in the future

niscinfo:
Hi, 
I have the same issue in v0.0.23. it ignores the sharedFileSystemPath parameter, and as nodePathMap is set to [] , the storageclass doesnt work. 
Thanks for the fix, hope it will be soon included. 

Regards,

PR #275, opened 2022-10-30
Fix sharedFilesystemPath not being available in helm deployment

meln5674:
Fixes https://github.com/rancher/local-path-provisioner/issues/274


derekbit:
Thanks for the fix. I will review later and release a new version for it by the end of this week.

derekbit:
LGTM. Thank you.

Issue #195, opened 2021-07-02
helper pod not able to set quotas


Elisabettaaa:
Hello ! i am following the example to enable quotas but everytime i got an error:

xfs_quotas: cannot setup path for mount /data:no such device or address

i am sure that the mounting point /data exists and it enables pquota.

derekbit:
Can you check the log and the error messages in the help pod?

liupeng0518:
same err.

PR #278, opened 2022-11-15
fix quota example

liupeng0518:
fix xfs quota example, the helper pod need mount /dev.
fix https://github.com/rancher/local-path-provisioner/issues/195


liupeng0518:
@derekbit

derekbit:
Will check it soon.

derekbit (approved):
LGTM, thank you.

PR #306, opened 2023-03-15
Multiple storage classes for same provisioner

samene:
We have multiple disks mounted on a node and depending on the type of storage requirement we want to create the local path pvc on a particular disk. Today we can specify multiple paths on a node via the `nodePathMap:` but the final path selection is random.

I propose to create multiple storage classes with a parameter of `path` that specifies the path to use. That way, using a storage class like below I can provision the storage on the right folder.

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ssd-local-path
  labels:
    app.kubernetes.io/name: local-path-provisioner
    helm.sh/chart: local-path-provisioner-0.0.23
    app.kubernetes.io/instance: local-path-provisioner
    app.kubernetes.io/version: "v0.0.23"
    app.kubernetes.io/managed-by: Helm
provisioner: cluster.local/local-path-provisioner
parameters:
  path: /data/ssd
volumeBindingMode: WaitForFirstConsumer
reclaimPolicy: Delete
```
If the path parameter is not found on the storage class, it falls back to existing behaviour.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -166,7 +166,7 @@ func (p *LocalPathProvisioner) watchAndRefreshConfig() {
 	}()
 }
 
-func (p *LocalPathProvisioner) getRandomPathOnNode(node string) (string, error) {
+func (p *LocalPathProvisioner) getPathOnNode(node string, requestedPath string) (string, error) {
 	p.configMutex.RLock()
 	defer p.configMutex.RUnlock()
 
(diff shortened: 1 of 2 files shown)


derekbit:
@samene 
In general, LGTM.
Can you use a more specific name rather than `path`?
In addition, please update the README.md for the new parameter as well. 
Thank you.

derekbit:
v0.0.24 will be released by this Wed. Put in v0.0.25 if can't catch up with v0.0.24.

provisioner.go
```diff
@@ -252,7 +253,13 @@ func (p *LocalPathProvisioner) Provision(ctx context.Context, opts pvController.
...
 		nodeName = node.Name
 	}
-	basePath, err := p.getRandomPathOnNode(nodeName)
+	var basePath string
+	basePath, err = p.getRandomPathOnNode(nodeName)
+	if storageClass.Parameters != nil {
+		if _, ok := storageClass.Parameters["path"]; ok {
+			basePath = storageClass.Parameters["path"]
+		}
+	}
```
derekbit, line 262: Need to check if `basePath` from the `storageClass.Parameters` is one of the `nodePathMap`.

derekbit:
> @derekbit I have updated the README.md. Can you suggest a better name for the `path` parameter?

@samene 
I don't have a better idea...`nodePath` might be more clear, WDYT?
BTW, can you help rebase and address the comment? I will releasev0.0.24 after getting this PR merged.
Thank you very much.

derekbit (approved):
LGTM. 
@samene Thanks for your contribution.

PR #311, opened 2023-03-20
Set default volume type in storage class

js185692:
We have  many `StatefulSet` workloads requiring the provisioner to create `local` volumes during provisioning. For several of these `StatefulSets`, we are unable to add the `volumeType` annotation which results in us having to use `hostPath` volumes which is unacceptable for us. Furthermore, it is tedious and error-prone to have to specify the `volumeType` annotation on every single PVC and would therefore prove useful to be able to specify a default volume type on the `StorageClass`.

I propose the addition of an annotation called `defaultVolumeType` on storage classes which will set the volume type for all volumes using that class. The annotation for the PVC can still be applied and will supersede the annotation on the storage class.

A basic example of a `StorageClass` with this annotation:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: local-path
  annotations:
    defaultVolumeType: local
    storageclass.kubernetes.io/is-default-class: true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ If the reload fails, the provisioner will log the error and **continue using the
 
 >time="2018-10-03T06:39:28Z" level=error msg="failed to load the new config file: config canonicalization failed: duplicate node yasker-lp-dev3"
 
-#### Volume Types
+### Volume Types
 
 To specify the type of volume you want the provisioner to create, add either of the following annotations;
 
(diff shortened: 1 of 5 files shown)


derekbit (commented):
In general, LGTM.

Suggest having a test case.

derekbit (approved):
LGTM.
Will test and merge it tomorrow.

derekbit:
@js185692 
Thanks for your contribution.

PR #317, opened 2023-03-23
Helm: Add the support of multiple storageclass configurations

### LINKED ISSUE #316: Helm: Allow to configure several storageclass
With the release of #306, it's now possible to use different storageclass to route the data to different directories. It's very useful :medal_sports:  but it's not yet possible to use in the helm chart.

WDYT to update it to support this kind of configuration:
```
## For creating the StorageClass automatically:
storageClasses:
  create: true

  ## Set a provisioner name. If unset, a name will be generated.
  # provisionerName: rancher.io/local-path

  classes:
    # The default storageclass with random path selection
    # For retro-compatibility with version < 0..0.24
    local-path:
      ## Set StorageClass as the default StorageClass
      ## Ignored if storageClass.create is false
      defaultClass: false
      ## ReclaimPolicy field of the class, which can be either Delete or Retain
      reclaimPolicy: Delete

      ## The path where to store the volumes
      ## Must exist in nodePathMap, if not specified, the path would be chosen
      ## randomly when provisioning.
      # nodePath: /my/path
```

vsellier:
Resolves #316

### DIFF
--- a/deploy/chart/local-path-provisioner/templates/NOTES.txt
+++ b/deploy/chart/local-path-provisioner/templates/NOTES.txt
@@ -1,5 +1,9 @@
 You can create a hostpath-backed persistent volume with a persistent volume claim like this:
 
+Possible class names supported by this deployment are {{ keys .Values.storageClasses.classes }}
+
+For example:
+
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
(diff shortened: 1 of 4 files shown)


JuniorJPDJ:
Having this would be amazing!

JuniorJPDJ:
This also adds an ability to select nodePath by helm which is the functionality I need now ;)

derekbit:
@vsellier 
Can you help resolve the conflicts? Thank you.

vsellier:
Hi,

If I'm not wrong, it seems the same  functionnality was already implemented and is available since the v0.0.27 (cf 1f742322165504ad304c8388f970c3f5a32994d9 and 90f926957798d96771945fb266af09e8d477b988), so, I guess this MR can be closed.

Feel free to reopen it if I'm wrong.

PR #319, opened 2023-04-02
Helm: Add common labels value and update uninstall for helm 3

kevinzwang:
A value that is often in Helm charts is the `commonLabels` value, which seems to be absent from this chart. Common labels are also added to each of the pods in the deployment in this PR. 

Additionally, I changed the readme install instructions as `helm uninstall` no longer has the `--purge` flag and instead purges by default.


derekbit (approved):
LGTM. Thanks for your contributions.

PR #321, opened 2023-04-06
Helm: add pod annotation, securityContext blocks

sergelogvinov:
Hello, thank you for this PVC driver.

This changes fix common pod-security warnings during deploy.


derekbit (approved):
LGTM

PR #325, opened 2023-04-16
Helm: add namespace

sergelogvinov:
Usually, helm uses in two cases: 
* helm update/install
* helm template

In first case, helm sets namespace during the deploy (everything fine here).
In second, the result of helm template can be used in `fluxcd` or `kubectl apply` pipeline.
But template does not have `namespace`, which can affect the result of deployment (deploy to default namespace)

Thank you.


derekbit (approved):
LGTM

PR #328, opened 2023-04-18
Explicitly list the container registry by default

jcpunk:
For my auditors, they want the registry listed for each container we are running.  Storing it here makes generating my reports much easier.

### DIFF
--- a/deploy/chart/local-path-provisioner/templates/configmap.yaml
+++ b/deploy/chart/local-path-provisioner/templates/configmap.yaml
@@ -29,6 +29,6 @@ data:
           {{- if .Values.privateRegistry.registryUrl }}
           image: {{ .Values.privateRegistry.registryUrl }}/{{ .Values.helperImage.repository }}:{{ .Values.helperImage.tag }}
           {{- else }}
-          image: {{ .Values.helperImage.repository }}:{{ .Values.helperImage.tag }}
+          image: {{ .Values.helperImage.defaultRegistry}}/{{ .Values.helperImage.repository }}:{{ .Values.helperImage.tag }}
           {{- end }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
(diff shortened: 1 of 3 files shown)


anothertobi:
This would introduce a breaking change for everyone currently configuring a registry as part of `.image.repository` like:
```
image:
  repository: registry.mycompany.com/rancher/local-path-provisioner
```

resulting in (broken):
```
image: "docker.io/registry.mycompany.com/rancher/local-path-provisioner:master-head"
```

We could use a backwards-compatible approach, like the rancher-system charts:
- https://github.com/rancher/system-charts/blob/dev-v2.7/charts/rancher-monitoring/v0.3.2/values.yaml#L384-L385
- https://github.com/rancher/system-charts/blob/dev-v2.7/charts/rancher-monitoring/v0.3.2/templates/_helpers.tpl#L126-L132
- https://github.com/rancher/system-charts/blob/dev-v2.7/charts/rancher-monitoring/v0.3.2/charts/prometheus/templates/prometheus.yaml#L87

However, this still requires you to provide custom values with the registry. I don't see a backwards-compatible way that would allow us to configure `docker.io` as the default registry in the Helm chart (non-empty default registry) since there are also some edge-cases where systems don't default to `docker.io` when no registry is given.

jcpunk:
That possibility non-defaulting to `docker.io` is what has our audit team concerned.  Without an explicit origin path they are uncomfortable.  I figured that discomfort might be shared by others and was hopeful for some sort of way clean up my site and provide something beneficial. 

I could switch the default registry to `None` and wrap it in logic so we don't get a weird leading `/` and keep `docker.io` in my local `values.yaml` but still track the image and tag from the chart.  Having the chart default to explicit would be my ideal, but sometimes life is like that.

derekbit:
@jcpunk The break change is a concern. Need to figure out a solution to it.

jcpunk:
I'll confess some interest in this still

jcpunk:
I'll confess some interest in this still

jcpunk:
I'll confess some interest in this still, any thoughts on how to proceed?

derekbit:
@jcpunk let's put it in the next release. Thanks.

jcpunk:
I'd still like to see some version of this.

PR #329, opened 2023-04-20
Trivy recommended hardening

jcpunk:
These hardening bits were recommended by the trivy scanner.

https://github.com/aquasecurity/trivy
https://github.com/aquasecurity/trivy-operator

### DIFF
--- a/deploy/chart/local-path-provisioner/values.yaml
+++ b/deploy/chart/local-path-provisioner/values.yaml
@@ -97,10 +97,10 @@ podAnnotations: {}
 podSecurityContext: {}
   # runAsNonRoot: true
 
-securityContext: {}
+securityContext:
+  seccompProfile:
+    type: RuntimeDefault
   # allowPrivilegeEscalation: false
-  # seccompProfile:
-  #   type: RuntimeDefault
   # capabilities:
   #   drop: ["ALL"]
   # runAsUser: 65534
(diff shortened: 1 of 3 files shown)


derekbit:
@jcpunk 
Thanks for your contribution.
Could you please link me to a reference for the feature? Thank you.

derekbit:
BTW, could you help resolve the conflict? Thank you.

derekbit:
Not sure if the hardening can introduce side effect?
I think we can gray the block out and leave a comment here. Users can enable it by updating the values.yaml.
@jcpunk WDYT?

jcpunk:
I'd still like to see some version of this.

derekbit:
Let's make it in v0.0.28.

jcpunk:
I'd still like to see some version of this.

deploy/chart/local-path-provisioner/values.yaml
```diff
@@ -74,10 +74,10 @@ podAnnotations: {}
 podSecurityContext: {}
   # runAsNonRoot: true
 
-securityContext: {}
+securityContext:
+  seccompProfile:
+    type: RuntimeDefault
```
derekbit, line 79: I'm wondering if we need to enable it by deafult. Or, users can enable it if they need it.

jcpunk: I would aim for by default and have folks loosen the security restrictions if they can't run with them.

jcpunk:
conflicts resolved

jcpunk:
I'd still like to see some version of this.

Issue #335, opened 2023-05-03
RBAC permissions constraint


sbocinec:
Hello local-path-provisioner community! 

Security scanners in our org complain about excessive RBAC permissions granted to `local-path-provisioner` through the [`local-path-provisioner-role` `ClusterRole`](https://github.com/rancher/local-path-provisioner/blob/master/deploy/chart/local-path-provisioner/templates/clusterrole.yaml), specifically [`*` granted for `["endpoints", "persistentvolumes", "pods"]`](https://github.com/rancher/local-path-provisioner/blob/master/deploy/chart/local-path-provisioner/templates/clusterrole.yaml#L12-L14). 

I wonder, does the provisioner indeed requires these excessive permissions? I'm currently checking the code to understand if we can constraint it in our deployment, though if anyone did something similar, I would appreciate if you can share your experience.

Thanks!

sbocinec:
I'm closing the issue as the PR #336 constraining the permissions has been already merge a year ago

PR #336, opened 2023-05-04
Constraint default RBAC permissions

sbocinec:
Constraint default set of RBAC permissions granted to `local-path-provisioner`:
* add namespace-scoped RBAC Role `local-path-provisioner-role`
* grant write permissions (create, delete, update, patch) for `pods` resource only in the Role (= granted only in the provisioner's namespace) as these should not be required globally. Remove these from the existing `ClusterRole`.
* remove  `endpoints` permission - I was not able to find any code reference working with `Endpoints`.
* do not use verbs wildcard for `persistentvolumeclaims` - this is to satisfy security scanners where wildcard is a no-no.

Motivation is to address #335 

I have tested the changes on multiple deployments, also going to intensively test in production in our managed solution.


derekbit (approved):
LGTM

PR #340, opened 2023-05-30
feat - Node name annotation

osherElm:
# Motivation
When a node is terminating, it is necessary to handle any leftover objects associated with that node. To accomplish this, the node name is passed as a parameter upon receiving the termination event. Currently, the code searches for a node name match within the affinity, which can be cumbersome. To simplify the filtering and handling process, I propose adding an annotation with the selected node name. This will allow for easier independent filtering and handling of objects using a separate controller.

Description of the change:
* Added the selected node name to the annotations.
* Removed redundant conversion operations.


derekbit (approved):
LGTM

PR #351, opened 2023-08-16
Allow parameters block for StorageClass

mdfaizsiddiqui:
Documentation around parameters block for storageClass - https://kubernetes.io/docs/concepts/storage/storage-classes/#parameters

For our use-case we wanted the volume to be of `xfs` file system and we were able to achieve it with this block -

```
parameters:
  fsType: xfs
```

So, I am adding a way in helm chart to specify parameters block for the storageClass.


derekbit:
@mdfaizsiddiqui 
local-path-provisioner actually creates a directory on a local filesystem (it can be xfs, ext4 and so one), so the `fsType` seems not really used.
Can you elaborate more on the `fsType`? Thank you.

JuniorJPDJ:
this is also useful if you want to achieve other things, like eg. limiting default storageClass to some notePath:

```
parameters:
  nodePath: /srv/k8s/pv/default
```

derekbit:
@mdfaizsiddiqui Can you help resolve the conflicts? Thank you.

mdfaizsiddiqui:
> @mdfaizsiddiqui Can you help resolve the conflicts? Thank you.

Hey, I've fixed all the conflicts, but due to this new change I had to moved the parameter introduced as part of this [PR](https://github.com/rancher/local-path-provisioner/commit/056e35801cee57f86c95b3211aa914a270dcb630)  to the new parameters section.

mdfaizsiddiqui:
@derekbit can you help merge this PR, thanks.

deploy/chart/local-path-provisioner/values.yaml
```diff
@@ -42,15 +42,17 @@ storageClass:
   ## Ignored if storageClass.create is false
   name: local-path
 
+  ## parameters configures the volumes belonging to the storage class
+  parameters: {}
+    # # Set a path pattern, if unset the default will be used
+    # pathPattern: "{{ .PVC.Namespace }}-{{ .PVC.Name }}"
+
```
derekbit, line 49: Can you fix the indent? Thank you.

PR #357, opened 2023-09-19
Set default volume type of storage class from Helm chart

nltimv:
This PR allows the user to set the default volume type of the storage class created by the Helm chart to either `local` or `hostPath` (default)


derekbit:
@nltimv Thanks for your contribution.

derekbit (approved):
LGTM

PR #358, opened 2023-09-29
deployment: Adapt values to be quoted when installed in templates

ardumont:
(initial title: deployment: Adapt 'workerThreads' to be a string)

There is a check in the main.go module [1] which expects it to be a string so it can convert it into an integer. Without this, we cannot actually provide the configuration without hitting a conversion error. This should fix it.

Currently, we are working around it by provided a double quoted string "'8'".

[1]  https://github.com/rancher/local-path-provisioner/blob/4d42c70e748fed13cd66f86656e909184a5b08d2/main.go#L254-L257

### DIFF
--- a/deploy/chart/local-path-provisioner/templates/deployment.yaml
+++ b/deploy/chart/local-path-provisioner/templates/deployment.yaml
@@ -58,15 +58,15 @@ spec:
             - {{ .Values.configmap.name }}
           {{- if .Values.workerThreads }}
             - --worker-threads
-            - {{ .Values.workerThreads }}
+            - {{ .Values.workerThreads | quote }}
           {{- end }}
           {{- if .Values.provisioningRetryCount }}
             - --provisioning-retry-count
-            - {{ .Values.provisioningRetryCount }}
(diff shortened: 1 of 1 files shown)


mirceanton:
I think the same problem is present for the `deletionRetryCount` and `provisioningRetryCount` fields as well. Maybe we should fix those in this PR as well?

derekbit:
> I think the same problem is present for the `deletionRetryCount` and `provisioningRetryCount` fields as well. Maybe we should fix those in this PR as well?

@mirceanton 
Agree.

@ardumont 
LGTM. Thanks for your contribution. Can you help fix `deletionRetryCount` and `provisioningRetryCount` as well? Many thanks.

ardumont:
> LGTM. Thanks for your contribution. Can you help fix deletionRetryCount and provisioningRetryCount as well? Many thanks.

Sure, done.
I've rebased the PR on the latest master too.

Cheers,

derekbit (approved):
LGTM. Thank you.

PR #360, opened 2023-10-15
feature: Add configuretion to run setup/teardown command in helper container instead of run with script

name212:
This pull request added two new configuration parameters (`setupCommand` and `teardownCommand`) for config data. Also passed `-a` argument with action type to script/command argument.

Goals.

I need to run setup and teardown action as binary because I need to run this binary in container image without base operating system (like alpine) for security reasons. This image does not contain shell and I are forced rewrite setup and teardown script in golang and pass it to image as binary. For this I need two configuration parameter and pass argument with action to binary.

I hope that pull request will be useful for community.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,9 @@ data:
                         "node":"yasker-lp-dev3",
                         "paths":[]
                 }
-                ]
+                ],
+                "setupCommand": "/manager",
+                "teardownCommand": "/manager"
         }
   setup: |-
         #!/bin/sh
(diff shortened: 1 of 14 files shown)


derekbit:
@name212 It would be nice if add the usage in the README.md. Thank you.

name212:
@derekbit 
Thank you for your response. I added information in to README file. Also I added example usage in examples/distroless directory.

derekbit (approved):
LGTM. Thanks for your contributions.

Issue #348, opened 2023-07-22
Allow multiple shared file system locations in sharedFileSystemPath.


heppstux:
Hi folks.

I'm wondering why I can not define multiple `sharedFileSystemPath`s analogous to the way I can configure `nodePathMap[*].paths`? This would allow to create and use storage classes (eg. the typical slow/fast setup.).

Are there any plans to implement this feature?

Thanks in advance.

Issue #38, opened 2019-08-07
creating multiple storage classes for multiple locations


ahmadalli:
I like to have some hostpaths (on all nodes) on `/mnt/b` and some on `/mnt/c`. is it possible for me to create multiple storage classes for multiple locations?

hdimitriou:
Just want to say that this is possible with the current setup. 
You just need a separate provisioner installation and you practically install teh chart 2 times with different config
```
storageClass:
  name: local-path-1
  provisionerName: rancher.io/local-path-1

nodePathMap: [
  {
    "node": "DEFAULT_PATH_FOR_NON_LISTED_NODES",
    "paths": ["/local-1"]
  }
]
```
and
```
storageClass:
  name: local-path-2
  provisionerName: rancher.io/local-path-2

nodePathMap: [
  {
    "node": "DEFAULT_PATH_FOR_NON_LISTED_NODES",
    "paths": ["/local-2"]
  }
]
```

yasker:
@hdimitriou ha, yeah. I was considering if you use the same provisioner name. Thanks for pointing this out!

ahmadalli:
@hdimitriou great! how does `provisionerName` get assigned? does it have numerical order or can it be customized?

yasker:
@ahmadalli You can set it manually using `--provisioner-name` in the command line, as a parameter for `local-path-provisioner`, e.g. after [`start`](https://github.com/rancher/local-path-provisioner/blob/master/deploy/local-path-storage.yaml#L68)

ahmadalli:
thanks. it worked :)
it would be good to add this to the documentation

dorsany:
so is it considered as a feature request to support 2 `storageClasses` in one `local-path-provisioner` controller to support multiple locations ?

PR #361, opened 2023-11-05
Feature/multiple storage classes

meln5674:
* Fixes https://github.com/rancher/local-path-provisioner/issues/38
* Fixes https://github.com/rancher/local-path-provisioner/issues/348

While is is currently possible to deploy multiple instances of the provisioner to support multiple storage classes, as well as to support both RWO, RWX, and ROX volumes at the same time, the additional resource consumption and deployment complexity are undesirable.

This patch adds an additional field to the `config.json` named `storageClassConfigs`, which is mutually exclusive with the existing `nodePathMap` and `sharedFilesystemPath` fields. This field is an object, whose keys are the names of storage classes, and its values are nested objects that contain either `nodePathMap` or `sharedFilesystemPath`. When a volume request is received, if this field is set, it is consulted for the configuration to use based on the name of the storage class, otherwise the original configuration fields are used.

This patch also updates the helm chart to add a matching option, which also deploys multiple storage class objects if selected.

This change is backwards compatible, installing this version of the provisioner should not break any existing installs.

One advantage of this change is that single-node k3s/k3d/kind clusters will be able to provide RWX and ROX storage classes out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,9 +198,11 @@ The helperPod is allowed to run on nodes experiencing disk pressure conditions, 
 `sharedFileSystemPath` allows the provisioner to use a filesystem that is mounted on all nodes at the same time.
 In this case all access modes are supported: `ReadWriteOnce`, `ReadOnlyMany` and `ReadWriteMany` for storage claims.
 
+`storageClassConfigs` is a map from storage class names to objects containing `nodePathMap` or `sharedFilesystemPath`, as described above.
+
 In addition `volumeBindingMode: Immediate` can be used in  StorageClass definition.
 
-Please note that `nodePathMap` and `sharedFileSystemPath` are mutually exclusive. If `sharedFileSystemPath` is used, then `nodePathMap` must be set to `[]`.
+Please note that `nodePathMap`, `sharedFileSystemPath`, and `storageClassConfigs` are mutually exclusive. If `sharedFileSystemPath` or `stroageClassConfigs` are used, then `nodePathMap` must be set to `[]`.
 
 The `setupCommand` and `teardownCommand` allow you to specify the path to binary files in helperPod that will be called when creating or deleting pvc respectively. This can be useful if you need to use distroless images for security reasons. See the examples/distroless directory for an example. A binary file can take the following parameters:
 | Parameter | Description |
--- a/provisioner.go
+++ b/provisioner.go
@@ -81,12 +81,22 @@ type NodePathMapData struct {
 	Paths []string `json:"paths,omitempty"`
 }
 
-type ConfigData struct {
+type StorageClassConfigData struct {
 	NodePathMap          []*NodePathMapData `json:"nodePathMap,omitempty"`
-	CmdTimeoutSeconds    int                `json:"cmdTimeoutSeconds,omitempty"`
 	SharedFileSystemPath string             `json:"sharedFileSystemPath,omitempty"`
-	SetupCommand         string             `json:"setupCommand,omitempty"`
-	TeardownCommand      string             `json:"teardownCommand,omitempty"`
+}
+
+type ConfigData struct {
+	CmdTimeoutSeconds int    `json:"cmdTimeoutSeconds,omitempty"`
+	SetupCommand      string `json:"setupCommand,omitempty"`
+	TeardownCommand   string `json:"teardownCommand,omitempty"`
+	StorageClassConfigData
+	StorageClassConfigs map[string]StorageClassConfigData `json:"storageClassConfigs"`
+}
+
+type StorageClassConfig struct {
+	NodePathMap          map[string]*NodePathMap
+	SharedFileSystemPath string
 }
 
 type NodePathMap struct {
@@ -94,11 +104,11 @@ type NodePathMap struct {
 }
 
 type Config struct {
-	NodePathMap          map[string]*NodePathMap
-	CmdTimeoutSeconds    int
-	SharedFileSystemPath string
-	SetupCommand         string
-	TeardownCommand      string
+	CmdTimeoutSeconds int
+	SetupCommand      string
+	TeardownCommand   string
+	StorageClassConfig
+	StorageClassConfigs map[string]StorageClassConfig
 }
 
 func NewProvisioner(ctx context.Context, kubeClient *clientset.Clientset,
@@ -177,7 +187,7 @@ func (p *LocalPathProvisioner) watchAndRefreshConfig() {
 	}()
 }
 
-func (p *LocalPathProvisioner) getPathOnNode(node string, requestedPath string) (string, error) {
+func (p *LocalPathProvisioner) getPathOnNode(node string, requestedPath string, c *StorageClassConfig) (string, error) {
 	p.configMutex.RLock()
 	defer p.configMutex.RUnlock()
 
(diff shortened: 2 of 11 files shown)


derekbit:
Can you help rebase the commit and update the README.md as well? Thank you.

meln5674:
Rebased. Can you elaborate on what sort of update to the README you're looking for? I did update the config.json definition section to include the new field I added.

FYSA I will be away from home for the next week, so if you'd like to put this on hold until then, I don't mind.

derekbit:
@meln5674 
Sorry, I missed the README.md is updated.
BTW, it would be nice if you would be able to add a test for the feature. Thank you.

meln5674:
Sorry for the delay on this, I ended up falling seriously ill after getting back from my trip.

I can't seem to find any information on how the tests here are supposed to work.

My default guess of `make test` shows `[no test files]`. I found `scripts/test` to be running `go test -cover -tags=test ./...`, which produces the same result. I noticed in `test/pod_test.go` its only being built when the `e2e` tag is present, so I tried `go test -cover -tags=e2e ./...`, which at least did something, but that produced the error `ERROR: a list of image names is required` which also happens on the master branch.

I did in fact write tests while writing this, though since I ran into the above issues since the start, I just wrote them in a shell script, which I've put into [this gist](https://gist.github.com/meln5674/2e39647d15e2ed99addb54ead0150ce1). If you give me any pointers on how to actually run the tests, as well as any on how best to integrate what I've written into the existing test setup, I'd really appreciate it.

timvahlbrock:
Hey @meln5674 , hope you are well again. I don't have any experience with Go and am still somewhat new to kubernetes, but as I'm looking forward to this feature I played around with the repo a bit. I checked how the CI is running the tests using the `.drone.yaml` file: It runs them using `dapper e2e-test`. I think dapper will be installed within the dir when executing the make file, so I ran `./.dapper e2e-test`. This actually runs a lot of tests (takes some time to get started though), but `TestPodWithRWOPVolume` is already failing on master in my case (even though CI is green). Hope this helps.

meln5674:
@timvahlbrock Thanks for the detective work, and I am in fact in better health again, thank you for your concern.

@derekbit I think I've worked out how the tests are structured now, and I've added a test which creates two storage classes, two matching PVC's, and mounts them to a single pod to match how the other tests work so I wouldn't have to re-work how the tests are structured. Let me know if you'd prefer I go back and do a 2-pod test instead.

timvahlbrock:
@meln5674 is this pr possibly not being reviewed because the pipeline is red (even though that's because it was killed...)?

meln5674:
I saw there were a few other PR builds that were killed, each because they never started their s390x build, so I assumed the project was working through some CI/CD issues.

timvahlbrock:
Can you set @derekbit as a reviewer explicitly or something? Or try to push something again to get the pipeline green? I don't wanna be too annoying or something, but other PRs seem to be getting merged.

timvahlbrock:
@meln5674 do you have permission to request a review on this PR?

derekbit:
> Can you set @derekbit as a reviewer explicitly or something? Or try to push something again to get the pipeline green? I don't wanna be too annoying or something, but other PRs seem to be getting merged.

@timvahlbrock Sorry for the late reply. Busy at Longhorn project.

> Or try to push something again to get the pipeline green

The error is due to that the s390x machine doesn't work.
amd64 and arm64 passed the tests. 
Are all updates ready to review? If yes, I will fix it and review the PR. soon.

meln5674:
Yes, everything should be ready for review.

timvahlbrock:
@derekbit  Slipped out of ToDo's again?

README.md
```diff
@@ -198,9 +198,11 @@ The helperPod is allowed to run on nodes experiencing disk pressure conditions,
 `sharedFileSystemPath` allows the provisioner to use a filesystem that is mounted on all nodes at the same time.
 In this case all access modes are supported: `ReadWriteOnce`, `ReadOnlyMany` and `ReadWriteMany` for storage claims.
 
+`storageClassConfigs` is a map from storage class names to objects containing `nodePathMap` or `sharedFilesystemPath`, as described above.
+
 In addition `volumeBindingMode: Immediate` can be used in  StorageClass definition.
 
-Please note that `nodePathMap` and `sharedFileSystemPath` are mutually exclusive. If `sharedFileSystemPath` is used, then `nodePathMap` must be set to `[]`.
+Please note that `nodePathMap`, `sharedFileSystemPath`, and `storageClassConfigs` are mutually exclusive. If `sharedFileSystemPath` or `stroageClassConfigs` are used, then `nodePathMap` must be set to `[]`.
```
derekbit, line 205: IIUC, the backward compatibility of an existing configmap with `nodePathMap` is also addressed, right?

meln5674: Yes. The ConfigMap under this patch can contain one and only one of nodePathMap, sharedFileSystemPath, or storageClassConfigs. An existing ConfigMap with one of nodePathMap and sharedFileSystemPath should continue to operate as-is, which I believe the tests demonstrate.

provisioner.go
```diff
@@ -246,11 +254,30 @@ func (p *LocalPathProvisioner) isSharedFilesystem() (bool, error) {
 	return false, fmt.Errorf("both nodePathMap and sharedFileSystemPath are unconfigured")
 }
 
+func (p *LocalPathProvisioner) pickConfig(storageClassName string) (*StorageClassConfig, error) {
+	if len(p.config.StorageClassConfigs) == 0 {
+		return &p.config.StorageClassConfig, nil
+	}
```
derekbit, line 260: This handles he `nodePathMap` as well as address the existing configmap format.

test/testdata/multiple-storage-classes/local-path-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+                                  "node":"DEFAULT_PATH_FOR_NON_LISTED_NODES",
+                                  "paths":["/opt/local-path-provisioner"]
+                          }
+                          ]
+                   },
+                   "local-path-shared": {
+                          "sharedFilesystemPath": "/opt/local-path-provisioner-shared"
+                   }
+            }
+    }
```
derekbit, line 22: https://raw.githubusercontent.com/rancher/local-path-provisioner/v0.0.26/deploy/local-path-storage.yaml using 
```
  config.json: |-
    {
            "nodePathMap":[
            {
                    "node":"DEFAULT_PATH_FOR_NON_LISTED_NODES",
                    "paths":["/opt/local-path-provisioner"]
            }
            ]
    }
```

Do you suggest using
```
  config.json: |-
    {
            "storageClassConfigs":{
                   "local-path": {
                          "nodePathMap": [
                          {
                                  "node":"DEFAULT_PATH_FOR_NON_LISTED_NODES",
                                  "paths":["/opt/local-path-provisioner"]
                          }
                          ]
                   }
            }
    }
```
is better?

The we can deprecate the old confg.

derekbit (approved):
The implementation LGTM.
Only one question for the deploy manifest.

test/testdata/multiple-storage-classes/local-path-config.yaml
meln5674: Not sure. It might make sense to leave it as-is so that existing installations don't change unexpectedly (even if the behavior is identical), and using the new config would be left as opt-in.

timvahlbrock:
Is there anything I can help with to progress this further?

derekbit:
@timvahlbrock
The PR looks good to me and is included in the upcoming v0.0.27. I will merge it after the broken CI is fixed. Thank you. :)

derekbit (approved):

JuniorJPDJ:
how nice, thanks! I'm waiting for this to be released!

arjunkanti:
When is v0.0.27 version getting released with this change..?

> @timvahlbrock The PR looks good to me and is included in the upcoming v0.0.27. I will merge it after the broken CI is fixed. Thank you. :)

PR #368, opened 2023-12-02
Repair code example in storageClass description.

c4lliope:
I copied this example and had a really rough go as a kubernetes beginner; I needed a custom storageClass for the RabbitMQ operator.

So my pods were broken for a long while, until I changed my storageClass to match the other `local-path` storageClass already enabled in my cluster. This `rancher.io/local-path` provisioner seems to be the key to an easy deployment of custom storageClasses.


derekbit (approved):
LGTM

PR #372, opened 2023-12-11
feat: add support to set folderName on PVC

vvanouytsel:
This makes it possible to set a 'folderName' annotation on a PVC. When set, the name of the created folder on disk will no longer be '$persistentVolumeName_$persistentVolumeClaimNamespace_$persistentVolumeClaimName'. Instead it will be overriden by the value of 'folderName' in the annotation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,16 @@ reclaimPolicy: Delete
 
 Here the provisioner will use the path `/data/ssd` when storage class `ssd-local-path` is used.
 
+### Changing the name of the PersistentVolume
+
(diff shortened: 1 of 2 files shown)


AlbanBedel:
The path must be validated otherwise it give access to the whole filesystem of the node.

XtraLarge:
> The path must be validated otherwise it give access to the whole filesystem of the node.

Yes, that's right. The (stiatic) basedir must be well validated and the (dynamic) part of the individual subdir-part under teh basedir must be unique, to be safe.

PR #375, opened 2023-12-28
Implement function to set HostPathType by annotation

kn3609571:
Implemented the function to set the PV hostPath type by annotation of StorageClass (defaultHostPathType) and PVC (hostPathType).
If neither is specified, DirectoryOrCreate is applied by default.

If the volume should not be provisioned when the parent directory does not exist, you can set the type of hostPath Directory via annotation above.

Refs rancher#137
Refs rancher#224


derekbit (commented):
@kn3609571 Can you update README.md and add an e2e test for the implementation? Thank you.

PR #377, opened 2024-01-05
Helm chart: Add support for helper tolerations

turip:
This is required if the current cluster uses tolerations for selecting certain node pools.

This change also includes the braking change to rename the .Values.helperImage to .values.helper.image, as I didn't wanted to introduce a helperTolerations top level variable.

If this is not fine, I can always revert that part of the change.


derekbit:
@turip 
Thanks for the feature. Introducing `helperTolerations` is fine for preventing the breaking change. Can you help address it? Thank you.

PR #382, opened 2024-01-23
adding pvc with node name example

sebastianohl:
If you want to use the local path provisioner in a multi node environment, it is required to explicitly set the node name via an annotation. Unfortunally, this looks very undocumented to me. I needed to pull the right annotation label from the source code.

As minimal solution, I extended the examples with an example with a node name.

An extended version cloud also mention this in the README.


derekbit (approved):

PR #388, opened 2024-03-11
Fix: Chart.yaml file is missing on helm install

### LINKED ISSUE #387: Helm: Chart.yaml file is missing on helm install
step to reproduce:
```
$ git clone https://github.com/rancher/local-path-provisioner.git
$ cd local-path-provisioner
$ helm install --name local-path-storage --namespace local-path-storage ./deploy/chart/
```

how to fix:
we should substitute `./deploy/chart/` with `./deploy/chart/local-path-provisioner`

jamshidi799:
Add `local-path-provisioner` to `./deploy/chart` in helm install commands

Fixes #387


derekbit (approved):

PR #393, opened 2024-03-22
Fix duplicate labels

runningman84:
fixes https://github.com/rancher/local-path-provisioner/issues/392

PR #394, opened 2024-03-29
Remove duplicate labels and add ability to set helperpod resource requests/limits

visokoo:
Piggybacking off this PR request from @runningman84: https://github.com/rancher/local-path-provisioner/pull/393 but retained the `selector.matchLabels` piece since it's not an immutable field after the release has been deployed and will actually block helm upgrades after the fact.

In addition to fixing the duplicate labels, this also adds the ability to add k8s resource requests/limits to the `busybox` helper-pod that's launched.

Tested this change locally in my environment and it's deploying and upgrading as expected with this flux `HelmRelease` manifest:

```
---
apiVersion: helm.toolkit.fluxcd.io/v2beta2
kind: HelmRelease
metadata:
  name: local-path-provisioner
  namespace: local-path-provisioner
spec:
  releaseName: local-path-provisioner
  interval: 10m
  chart:
    spec:
      chart: ./deploy/chart/local-path-provisioner
      version: 0.0.25-dev
      sourceRef:
        kind: GitRepository
        name: local-path-provisioner
        namespace: local-path-provisioner
      reconcileStrategy: Revision
  values:
    storageClass:
      provisionerName: local-path
    resources:
      requests:
        memory: 150Mi
        cpu: 60m
      limits:
        memory: 300Mi
        cpu: 120m
    helperPod:
      resources:
        requests:
          memory: 100Mi
          cpu: 10m
        limits:
          memory: 200Mi
          cpu: 20m

    # Number of retries of failed volume provisioning. 0 means retry indefinitely.
    provisioningRetryCount: 5
    # Number of retries of failed volume deletion. 0 means retry indefinitely.
    deletionRetryCount: 5
```

### DIFF
--- a/deploy/chart/local-path-provisioner/templates/configmap.yaml
+++ b/deploy/chart/local-path-provisioner/templates/configmap.yaml
@@ -38,3 +38,6 @@ data:
           image: {{ .Values.helperImage.repository }}:{{ .Values.helperImage.tag }}
           {{- end }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+      resources:
+        {{- toYaml .Values.helperPod.resources | nindent 8 }}
+
--- a/deploy/chart/local-path-provisioner/templates/configmap.yaml
+++ b/deploy/chart/local-path-provisioner/templates/configmap.yaml
@@ -39,5 +39,5 @@ data:
           {{- end }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           resources:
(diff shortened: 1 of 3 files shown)


derekbit (approved):
LGTM. Thanks for the contribution.

visokoo:
@derekbit, anything else I need to do here or are we good to merge?

derekbit:
> @derekbit, anything else I need to do here or are we good to merge?

No. I've merged it. It will be included in v0.0.27 released at the end of this May. Thank you.

PR #399, opened 2024-05-09
Automatic reloading of the helper pod manifest by the provisioner

js185692:
# About

In the readme, there is a section on _reloading_ where it discusses the ability of the provisioner to be able to automatically pick up changes made to the config.

> The provisioner supports automatic configuration reloading. Users can change the configuration using kubectl apply or kubectl edit with config map local-path-config. There is a delay between when the user updates the config map and the provisioner picking it up.

I have found that, while this works for changes made to the `setup` and `teardown` scripts, when I make an update to the helper pod manifest in the config map the provisioner does not pick it up. I found that I would have to restart the provisioner for the helper pod changes to take effect.

# Changes

I have implemented a watch function of the helper pod manifest, that takes the file at the mount location (`/etc/config/`) and sets the provisioner field of `helperPod` to that of the contents of the file.

I have included a new env variable for the provisioner container that specifies the location of the config. This implementation is backwards compatible with deployments that don't have this env variable defined, as the resulting behaviour will be that of the current provisioner; meaning that changes made to the helper pod manifest in the config map will only take effect when the provisioner is restarted.

**Note:** I also updated in util to use `io.ReadAll` instead of `ioutil.ReadAll` as it was deprecated. I can change this back if it's necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,12 @@ The scripts receive their input as environment variables:
 
 #### Reloading
 
-The provisioner supports automatic configuration reloading. Users can change the configuration using `kubectl apply` or `kubectl edit` with config map `local-path-config`. There is a delay between when the user updates the config map and the provisioner picking it up.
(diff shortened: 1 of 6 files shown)


derekbit:
@js185692 Thanks for the contribution. Ping me if it is ready. Thank you.

js185692:
@derekbit the PR should be ready for review now, thanks!

derekbit (approved):

PR #385, opened 2024-02-12
Add support for custom path patterns

AlbanBedel:
This PR implement support for configuring the name of the created volume directories using a template as requested in #302.

PR #400, opened 2024-05-14
Allow for the configuration of shortened folder names

js185692:
# Context

I use the local storage provisioner with LVM, and I have encountered a few times now where the folder being created is exceeding the [Logical Volume name length limit](https://github.com/lvmteam/lvm2/blob/25a87ea16aba335b5f33621cd5ac6ff651f0bbb1/tools/lvrename.c#L177C11-L177C20). I would like to introduce a new boolean field in the config for the provisioner called `shortenedFolderName` that, when set, creates the folder with the same name as that of the PV (aka `pvc-<uuid>`).

The default naming convention will still be the same as the current implementation (that being `<PV name>-<PVC namespace>-<PVC name>`).


js185692:
Closing in favour of the work done in #385.

PR #402, opened 2024-05-22
Give the helper pod more range of MCS categories

galal-hussein:
- https://github.com/k3s-io/k3s/issues/10130
- 
The fix will provide the helper pod with more MCS categories which will fix the issue with having two pods trying to access shared files (the helper setup pod, and the helper delete pod)


brandond (approved):

derekbit:
@galal-hussein 
Can you give us more context and purpose of the PR? Thank you.

galal-hussein:
@derekbit I have linked the related issue which this PR solves, you can also check out my latest comment https://github.com/k3s-io/k3s/issues/10130#issuecomment-2125727808 it describes the problem and the solution in the PR

derekbit (commented):

derekbit (approved):

PR #403, opened 2024-05-24
Migrate CI to github Actions

mantissahz:
~`e2e-test` is not included because there are many errors in my local testing.~

We will have the step `make e2e-test` after the step `make ci`.

cc @innobead, @bk201, @derekbit, @FrankYang0529


.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+    # For multi-platform support
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Declare Branch
+      run: |
+        echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> "$GITHUB_ENV"
+    - name: Login to Docker Hub
```
FrankYang0529, line 65: In rancher organization, we may need to read docker credential like this.

https://github.com/harvester/load-balancer-harvester/blob/0025da293f105cc500a27795249e4f952d80adcd/.github/workflows/template-build.yml#L19-L21

https://github.com/harvester/load-balancer-harvester/blob/0025da293f105cc500a27795249e4f952d80adcd/.github/workflows/template-build.yml#L35-L41

FrankYang0529 (approved):
Overall LGTM. For pushing image, we can merge first and fix it if it can't work.

derekbit:
I will merge it after releasing v0.0.27.

derekbit:
@mantissahz 
Can you build images for arm and amd64 platforms?

mantissahz:
@derekbit,

I can build arm64 and amd64 images as https://hub.docker.com/repository/docker/jamesluhz/local-path-provisioner/tags.

Do we also need the arm image (it is built in Drone CI)?

derekbit:
> @derekbit,
> 
> I can build arm64 and amd64 images as https://hub.docker.com/repository/docker/jamesluhz/local-path-provisioner/tags.
> 
> Do we also need the arm image (it is built in Drone CI)?

Yes, need `arm` as well.

mantissahz:
https://hub.docker.com/repository/docker/jamesluhz/local-path-provisioner/tags
Images built with arch `arm64`, `arm`, `amd64`

.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+    - name: Checkout Codes
+      uses: actions/checkout@v4
+
+    # Build binaries
+    - name: Run CI
+      run: make ci
+
+    # Run e2e test
+    #- name: Run E2e Test
+    #  run: make e2e-test
```
derekbit, line 24: We can add `make e2e-test` now. I will fix the flaky issue later.

mantissahz: Sure. Added.

.github/workflows/build.yml
mantissahz: Modified. Thanks

derekbit (approved):

Issue #413, opened 2024-06-07
provisioner assumes that a node's metadata.name == metadata.labels["kubernetes.io/hostname"]


jan-g:
As per https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#built-in-node-labels - 

[[[
The value of these labels is cloud provider specific and is not guaranteed to be reliable. For example, the value of kubernetes.io/hostname may be the same as the node name in some environments and a different value in other environments.
]]]

Assumptions around these being equal have caused related bugs with k8s itself, see: https://github.com/kubernetes/kubernetes/issues/125336

For non-shared nodes, using a MatchFields entry instead on `metav1.ObjectNameField` may prove more robust.

The main visible consequence of this is that the delete helper pod can't be scheduled, for clouds where metadata.name of a node doesn't align with its hostname label.

In terms of forward/backward compatibility, this code in `createHelperPod`

```
	if o.Node != "" {
		helperPod.Spec.NodeName = o.Node
	}
```

probably needs to distinguish between having a node name and having a selector (and just replicating the selector in the latter case, letting the scheduler find the right spot to place the pod).

jan-g:
Something like #414 should do the right thing I think.

jan-g:
PR is available for this.

PR #414, opened 2024-06-07
Bugfix: pvs not deleting was: Remove the assumption that a node's name == its hostname

jan-g:
As per https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#built-in-node-labels

[[[
Note:
The value of these labels is cloud provider specific and is not guaranteed to be reliable. For example, the value of kubernetes.io/hostname may be the same as the node name in some environments and a different value in other environments.
]]]

(This is a quick handwave in the direction of #413)

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -422,20 +422,16 @@ func (p *LocalPathProvisioner) provisionFor(opts pvController.ProvisionOptions, 
 		// affinity, as path is accessible from any node
 		nodeAffinity = nil
 	} else {
-		valueNode, ok := node.GetLabels()[KeyNode]
-		if !ok {
-			valueNode = nodeName
-		}
 		nodeAffinity = &v1.VolumeNodeAffinity{
 			Required: &v1.NodeSelector{
 				NodeSelectorTerms: []v1.NodeSelectorTerm{
 					{
-						MatchExpressions: []v1.NodeSelectorRequirement{
+						MatchFields: []v1.NodeSelectorRequirement{
 							{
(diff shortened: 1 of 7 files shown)


RafalKorepta (approved):
LGTM

jan-g:
I'm not a kustomize expert - I think it'd be useful to demonstrate this working both "from scratch" and also with a granfathered-in PV with the legacy hostname affinity requirement on it - could do with a hand there! (Hence the draft status)

jan-g:
This could really do with https://github.com/rancher/local-path-provisioner/pull/416 to close down the raciness of the e2e tests.

jan-g:
Fixed up the new e2e test (it creates a PV that should be cleaned up & waits on the provisioner; therefore we don't want deleteKustomizeDeployment to target it)

jan-g:
Commenting ro remove the stale tag

liangjuf:
Hi @jan-g thanks for the great fix, when can we get this merged to master? @derekbit @RafalKorepta

RafalKorepta:
Hi @liangjuf I'm not maintainer, so I don't have power to merge.

jan-g:
There's this PR and the one next to it that fixes up some issues with occasional e2e test failures (that means this one's tests should work reliably). This'll need rebasing atop the e2e one once that lands.

Maintainers are a little thin on the ground (and derekbit is in a different tz) so I'd ask for a little patience.

derekbit (approved):
LGTM

PR #416, opened 2024-06-07
e2e: wait for the provisioner to clean up PV resources

jan-g:
If we're unlucky, the local-path-provisioner can be killed before it has a chance to clean up orphened PVs. These may then hang around while subsequent tests run. Depending on the lexical sort order of those PVs, the `typeCheckCmd` in `runTest` may select a moribund PV, causing the test to randomly fail.

### DIFF
--- a/test/pod_test.go
+++ b/test/pod_test.go
@@ -39,6 +39,8 @@ func (p *PodTestSuite) SetupSuite() {
 	cmds := []string{
 		fmt.Sprintf("kind create cluster --config=%s --wait=120s", testdataFile("kind-cluster.yaml")),
 		fmt.Sprintf("kind load docker-image %s", p.config.IMAGE),
+		// We'll be replacing this, but applying particular labels to it.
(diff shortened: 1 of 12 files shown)


derekbit:
The flaky failure condition seems related the cleanup order of the storageclass.

jan-g:
It's just passing the output of kustomize into kubectl delete. Every kustomize call is (re)installing the storageclass helper.

This PR explicitly tags the storageclass stuff so that kubectl delete has a label to *only* ditch the pod & pvc before it waits on the cleanup completing.

jan-g:
@derekbit is there more required to progress this?

jan-g:
@derekbit I cannot merge this - can you do the honours please?

derekbit:
> @derekbit I cannot merge this - can you do the honours please?

Sure. I'm reviewing it. No worries. Will merger it if there is no concern. Thanks you.

derekbit:
@jan-g Could you resolve the conflicts? Thank you.

derekbit (approved):
LGTM

PR #419, opened 2024-06-10
Add support for Block devices

Deltachaos:
Following up to #125 and after #127 has been merged, it should be up to the helper Pod to decide up on support for Block devices or not.

In fact i have it running with the scripts from #125

But to create the ressources the PV that gets created needs to inherit the correct volumeMode and block support should be allowed.


derekbit:
@Deltachaos 
Can you add an e2e test case and an example for the feature? Thank you.

derekbit:
ping @Deltachaos

Deltachaos:
I will try to do one now

derekbit:
Let's put in the v0.0.30.

Issue #436, opened 2024-07-29
invalid chart reference: validation: chart.metadata.version "master-head" is invalid


applike-ss:
When deploying the helm chart via flux2, i do get the error: `invalid chart reference: validation: chart.metadata.version "master-head" is invalid`.

Can/Should we use a proper version scheme here, like SemVer?

derekbit:
Sure.
BTW, can you switch to v0.0.28 instead? master-head is only for development purpose.

applike-ss:
I've opened a PR here: https://github.com/rancher/local-path-provisioner/pull/437

applike-ss:
I switched to referring to the tagged version instead in the GitRepository resource of flux2, closing

PR #437, opened 2024-07-29
chart: fix helm chart version + appVersion reference

applike-ss:
Fixes: https://github.com/rancher/local-path-provisioner/issues/436

i've omitted the "v" in version, as it would not be valid SemVer 2.

### DIFF
--- a/deploy/chart/local-path-provisioner/values.yaml
+++ b/deploy/chart/local-path-provisioner/values.yaml
@@ -5,7 +5,7 @@ commonLabels: {}
 
 image:
   repository: rancher/local-path-provisioner
-  tag: master-head
+  tag: v0.0.28
   pullPolicy: IfNotPresent
 
 helperImage:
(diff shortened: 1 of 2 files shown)


deploy/chart/local-path-provisioner/values.yaml
```diff
@@ -5,7 +5,7 @@ commonLabels: {}
 
 image:
   repository: rancher/local-path-provisioner
-  tag: master-head
+  tag: v0.0.28
```
derekbit, line 8: Can you update it to v0.0.28-dev?

applike-ss: At least for me, this tag does not exist:
```
$ docker pull rancher/local-path-provisioner:v0.0.28-dev

What's next:
    View a summary of image vulnerabilities and recommendations → docker scout quickview rancher/local-path-provisioner:v0.0.28-dev
Error response from daemon: manifest for rancher/local-path-provisioner:v0.0.28-dev not found: manifest unknown: manifest unknown
```

Are you sure the `-dev` suffix is correct? I only see an rc1 tag on docker hub.

derekbit: Why not switch to https://github.com/rancher/local-path-provisioner/tree/v0.0.28 for the flux2?

applike-ss: I just checked back and you are right, i should switch to just using the tag.

/close

Issue #454, opened 2024-09-25
Node Affinity not set on v0.0.29


msander:
I am using the local-path-provisioner with the volumeType "local". Since 0.0.29 I have the problem that the node affinity of the PV is not set. In 0.0.28 it worked correctly.
I am using rke2 (1.30.4). The PV is created as follows:

```
root@vm1:~# kubectl describe pv pvc-8b95ded6-dbc9-4e6c-8b7b-531c62ba28b6
Name:              pvc-8b95ded6-dbc9-4e6c-8b7b-531c62ba28b6
Labels:            <none>
Annotations:       local.path.provisioner/selected-node: vm1
                   pv.kubernetes.io/provisioned-by: rancher.io/local-path
Finalizers:        [kubernetes.io/pv-protection]
StorageClass:      local-path
Status:            Bound
Claim:             default/cnpg-database-1
Reclaim Policy:    Delete
Access Modes:      RWO
VolumeMode:        Filesystem
Capacity:          10Gi
Node Affinity:     
  Required Terms:  
    Term 0:        <none>
Message:           
Source:
    Type:  LocalVolume (a persistent volume backed by local storage on a node)
    Path:  /opt/local-path-provisioner/pvc-8b95ded6-dbc9-4e6c-8b7b-531c62ba28b6_default_cnpg-database-1
Events:    <none>
```

When going back to 0.0.28, the node affinity gets set correctly.

This is my storage class

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: local-path
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
    defaultVolumeType: local
provisioner: rancher.io/local-path
volumeBindingMode: WaitForFirstConsumer
```

PVC
```
Name:          cnpg-database-1
Namespace:     default
StorageClass:  local-path
Status:        Bound
Volume:        pvc-8b95ded6-dbc9-4e6c-8b7b-531c62ba28b6
Labels:        cnpg.io/cluster=cnpg-database
               cnpg.io/instanceName=cnpg-database-1
               cnpg.io/pvcRole=PG_DATA
Annotations:   cnpg.io/nodeSerial: 1
               cnpg.io/operatorVersion: 1.24.0
               cnpg.io/pvcStatus: ready
               pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: rancher.io/local-path
[...]

Issue #451, opened 2024-09-04
PV nodeAffinity.required.nodeSelectorTerms matchFields does not result in the kube scheduler placing pods on the same node as the PV


kmurray01:
Observing this on the latest v0.0.29 release and master-head local-path-provisoner on K3s v1.30.4+k3s1.

PVs created with the new local-path-provisoner image have an updated `nodeAffinity.required.nodeSelectorTerms` as depicted below

```
nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchFields:
        - key: metadata.name
          operator: In
          values:
          - my-agent-host.example.com
```

This change originated from PR https://github.com/rancher/local-path-provisioner/pull/414, that was subsequently included into the latest v0.0.29 release and master-head.

Deploying a pod specifying a `persistentVolumeClaim` to an associated pvc, the pod is scheduled on a different node, not `my-agent-host.example.com`.  That pod then fails to initialize as it's unable to mount the PV volume path on `my-agent-host.example.com`. 

Previously on v0.0.28, the PV `nodeAffinity.required.nodeSelectorTerms` is as below and this works. The kube scheduler places the pod on the same node on which the PV local path volume is created, i.e. `my-agent-host.example.com` in this example. 

```
nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - my-agent-host.example.com
```

It would seem switching the `nodeAffinity.required.nodeSelectorTerms` to `matchFields` for node field `metadata.name` does not work or the kube scheduler does not comply with that nodeAffinity. 

Also it's important to highlight on the K3s node, the value of `metadata.name` matches `my-agent-host.example.com`.

kmurray01:
@derekbit @jan-g if you could please triage

haojingcn:
yes, i met same issue.

sts with volumePersistentTemplate, generate the pv yaml like
![image](https://github.com/user-attachments/assets/ecd1808f-6d9d-4ce0-a6bc-522e50015b74)

metadata.name is not nested label with K8s(1.19-1.26)

huangguoqiang:
i met same issue.

spirkaa:
Same problem. After draining node pod will be scheduled on another node without PV resulting in crashloop or another incorrect behavior of app inside.

PR #458, opened 2024-10-05
Revert "Remove the assumption that a node's name == its hostname"

derekbit:
Revert "Remove the assumption that a node's name == its hostname" due to the compatibility issues reported by https://github.com/rancher/local-path-provisioner/issues/454 and https://github.com/rancher/local-path-provisioner/issues/451


jan-g:
Just catching up with this; email's been going into a black hole.

I think the issue revolves around the k8s scheduler itself - looks to me like the PRs to fix the assumptions there were not completed.

Issue #473, opened 2025-01-10
Helm chart supports using namespaceOverride.


ace0527:
Currently, the `namespace` attribute value of all namespace-related components in the templates is `namespace: {{ .Release.Namespace }}`, making it impossible to use `namespaceOverride` in dependencies to define the namespace.
A more flexible and compatible implementation should be as follows:
Add a template:
```yaml
{{- define "local-path-provider.namespace" -}}
{{- if .Values.namespaceOverride }}
{{- .Values.namespaceOverride }}
{{- else }}
{{- .Release.Namespace }}
{{- end }}
{{- end -}}
```
Assign the namespace to the component:
```yaml
namespace: {{ include "phpmyadmin.namespace" . }}
```
Currently, to modify the namespace, you can only pull the dependency locally and modify the code.
I hope future versions can be compatible with `namespaceOverride`.

derekbit:
@tzabbi I will review it. Thanks for the contribution.

PR #478, opened 2025-02-11
feat: add ability to set custom namespace

tzabbi:


tzabbi:
closes #473

deploy/chart/local-path-provisioner/templates/deployment.yaml
```diff
@@ -72,7 +72,7 @@ spec:
               mountPath: /etc/config/
           env:
             - name: POD_NAMESPACE
-              value: {{ .Release.Namespace }}
+              value: {{ include "local-path-provisioner.namespace" . }}
```
Skaronator, line 75: Can be changed to be even more dynamci:

```suggestion
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
```

tzabbi: good point will implement it. Cannot add your suggestion so I make another commit.

tzabbi:
@derekbit do you have some update regarding your review timeslot? :)

derekbit:
> @derekbit do you have some update regarding your review timeslot? :)

Looks good to me. I will test and merge it by the end of this week. Thank you @tzabbi

tzabbi:
Thank you very much! Thanks for your time I appreciate it!

deploy/chart/local-path-provisioner/values.yaml
```diff
@@ -2,6 +2,7 @@
 
 replicaCount: 1
 commonLabels: {}
+namespace: local-path-provisioner
```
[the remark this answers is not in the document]
tzabbi: Yes thanks, I wonder how I missed this.

derekbit (approved):
LGTM

PR #480, opened 2025-02-27
fix: don't try to clean up pvs on nodes that are gone

marcusramberg:
We're running local-provisioner to provide local storage for CI runners where nodes come and go pretty frequently.
We observe that the provisioner is trying run clean up on nodes that are already gone, which causes helper pods
to be stuck in pending state as they cannot be scheduled.

This PR adds a check to see if the node still exists before trying to clean up the node.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -486,6 +486,11 @@ func (p *LocalPathProvisioner) deleteFor(pv *v1.PersistentVolume, c *StorageClas
 		if node == "" {
 			logrus.Infof("Deleting volume %v at %v", pv.Name, path)
 		} else {
+			// Check if node still exists
+			if _, err := p.kubeClient.CoreV1().Nodes().Get(context.TODO(), node, metav1.GetOptions{}); err != nil && k8serror.IsNotFound(err) {
+				logrus.Infof("Node %v does not exist, skipping cleanup of volume %v", node, pv.Name)
+				return nil
+			}
 			logrus.Infof("Deleting volume %v at %v:%v", pv.Name, node, path)
 		}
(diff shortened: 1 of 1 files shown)


marcusramberg:
@derekbit Thoughts about this PR? We're running it in production from a fork now and it has resolved our issue of stuck pvs from old nodes and stuck helper prs trying to schedule on non-existing nodes. I guess it would also address the issues you're seeing in #416 with stuck pvs from previous runs?

provisioner.go
```diff
@@ -483,6 +483,11 @@ func (p *LocalPathProvisioner) deleteFor(pv *v1.PersistentVolume, c *StorageClas
 		if node == "" {
 			logrus.Infof("Deleting volume %v at %v", pv.Name, path)
 		} else {
+			// Check if node still exists
+			if _, err := p.kubeClient.CoreV1().Nodes().Get(context.TODO(), node, metav1.GetOptions{}); err != nil {
+				logrus.Infof("Node %v does not exist, skipping cleanup of volume %v", node, pv.Name)
+				return nil
+			}
```
derekbit, line 490: @marcusramberg WDYT?
```suggestion
			if _, err := p.kubeClient.CoreV1().Nodes().Get(context.TODO(), node, metav1.GetOptions{}); err != nil && apierrors.IsNotFound(err) {
				logrus.Infof("Node %v does not exist, skipping cleanup of volume %v", node, pv.Name)
				return nil
			}
```

derekbit (approved):
LGTM @marcusramberg. Thanks for your contribution. The improvement will be in v0.0.33 that is scheduled in Oct

PR #486, opened 2025-04-11
Fix helper pod tolerations

sbocinec:
Currently, helperPod tolerates any taints as the provisioner always adds `- operator: Exists` toleration to the helperPod spec. This toleration is extremely permissive, defeating the purpose of setting extra helperPod tolerations through the config. 

This PR address the issue with the following:
1. Do not apply the default tolerations if helperPod tolerations are provided through the config
2. Replace the extremely permissive toleration with:
```
tolerations:
  - key: node.kubernetes.io/disk-pressure
    operator: Exists
    effect: NoSchedule
```

that's already used across majority of the helper-pod examples, assuming it's as a sane new default.

Fixes https://github.com/rancher/local-path-provisioner/issues/487

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -602,7 +602,9 @@ func (p *LocalPathProvisioner) createHelperPod(action ActionType, cmd []string, 
 	}
 	lpvTolerations := []v1.Toleration{
 		{
+			Key:      v1.TaintNodeDiskPressure,
 			Operator: v1.TolerationOpExists,
+			Effect:   v1.TaintEffectNoSchedule,
 		},
 	}
 	helperPod := p.helperPod.DeepCopy()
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

derekbit:
Hello @sbocinec, why is the adjustment only applied to disk-pressure?

sbocinec:
@derekbit working on the PR I'd analyzed the code and understood, local-path-provisioner always adds only the `disk-pressure` taint that's considered as a default:
* https://github.com/rancher/local-path-provisioner/blob/727d9b2146ca927571c39df0c99aba274f09834c/deploy/chart/local-path-provisioner/templates/configmap.yaml#L49-L51
* https://github.com/rancher/local-path-provisioner/blob/727d9b2146ca927571c39df0c99aba274f09834c/deploy/local-path-storage.yaml#L155-L157
* https://github.com/rancher/local-path-provisioner/blob/727d9b2146ca927571c39df0c99aba274f09834c/deploy/example-config.yaml#L40-L42

This change is aligned with all the above deployment defaults. Users are still allowed to override the defaults by providiing their own helper-pod spec.

Issue #89, opened 2020-06-05
Publish Helm chart in a chart repository to simplify installation


chgl:
Having to download the repository to install the provisioner via the helm chart is a bit more inconvenient than if it were available via a chart repository. A central repository for all Rancher charts could be hosted directly via GitHub Pages (see https://medium.com/@mattiaperi/create-a-public-helm-chart-repository-with-github-pages-49b180dbb417).

I see that there already is a GitHub repo (https://github.com/rancher/charts), but I can't find any reference to a corresponding Helm chart repo.

appanycd:
+1

yasker:
We're working on adding local path provisioner to a Helm repo.

sergeyshaykhullin:
+1

ribx:
https://github.com/rancher/local-path-provisioner/tree/master/deploy/chart

softwareplumber:
+1

sergeyshaykhullin:
@yasker Hello, any updates?

will-beta:
expecting updates

sergeyshaykhullin:
That is a prerequisite to add helm repo support?

rmoreas:
[helm/chart-releaser-action](https://github.com/helm/chart-releaser-action) can help to publish charts on GitHub Pages.

sergeyshaykhullin:
Also you can publish chart in OCI format https://github.com/appany/helm-oci-chart-releaser

sergeyshaykhullin:
I have found that @ebrianne published copy of this chart and it is available on:
- Artifacthub https://artifacthub.io/packages/helm/ebrianne/local-path-provisioner
- GitHub https://github.com/ebrianne/helm-charts/tree/master/charts/local-path-provisioner
- GitHub Pages https://ebrianne.github.io/helm-charts

```
helm repo add ebrianne https://ebrianne.github.io/helm-charts
helm install my-local-path-provisioner ebrianne/local-path-provisioner --version 0.0.19 
```

For me it is sutable for temporary workaround 🎉

ebrianne:
@sergeyshaykhullin I did it because I saw that no chart was available for it. Pretty simple and working good for my needs.

Forbzy:
Im glad that https://charts.containeroo.ch and https://ebrianne.github.io/helm-charts. Will these be kept up to date with new releases? Currently version 0.0.20 is out. Any word on when rancher will do there own one?

ebrianne:
[...]

PR #501, opened 2025-06-02
OCI  based helm chart build workflow added

supu2:
This PR container OCI based helm chart distribute workflow.
#89  Issue can be closed.
```
# Debug
helm template -n kube-system local-path-provisioner oci://ghcr.io/rancher/local-path-provisioner/charts/local-path-provisioner --version 0.0.31
# Deploy
helm upgrade --install -n kube-system local-path-provisioner oci://ghcr.io/rancher/local-path-provisioner/charts/local-path-provisioner --version 0.0.31
```

Link: https://helm.sh/docs/topics/registries/

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -1,0 +1,31 @@
+name: build
+on:
+  push:
+    branches:
+    - master
+    - v*
+    tags:
+    - v*
+  pull_request:
+env:
+  HELM_REGISTRY: ghcr.io
+  CHART_DIR: deploy/chart/local-path-provisioner
+jobs:  
+  build_push_helm_chart:
+    name: Build and Push Chart
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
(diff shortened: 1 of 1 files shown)


euven:
@derekbit I cannot see how straight up publishing this to GHCR in the same repo is any different from having the code (and chart) available here in github. It just gives users another way of "downloading" the chart code.

Could you merge and publish, then revert if Rancher is not happy?

Having to use third-party helm repos do not sit right and I'm sure Rancher will agree that this is a great first step in making the chart more generally available.

derekbit:
> @derekbit I cannot see how straight up publishing this to GHCR in the same repo is any different from having the code (and chart) available here in github. It just gives users another way of "downloading" the chart code.
> 
> Could you merge and publish, then revert if Rancher is not happy?
> 
> Having to use third-party helm repos do not sit right and I'm sure Rancher will agree that this is a great first step in making the chart more generally available.

Unfortunately, the secret GITHUB_TOKEN is not set. I've filed the request to rancher. Let's wait for response.

derekbit:
Dockerhub also supports OCI artifacts. Do you consider improve this part as well?
https://docs.docker.com/docker-hub/repos/manage/hub-images/oci-artifacts/

derekbit (approved):
Let's try it. Thanks.

PR #502, opened 2025-06-08
use existing kubeClient when saving helper logs

sceneryback:
when saving helper logs, there is no need to create a new kubeClient. Just use the existing one in provisioner


sceneryback:
Hi @derekbit please take a look

derekbit (approved):
LGTM. Thanks for contribution.

Issue #417, opened 2024-06-07
Client-side throttling is logged. Any configuration options to adjust?


pp-akursar:
We're seeing lots of client-side throttling logs, for example:

```
I0607 16:06:28.256034       1 request.go:697] Waited for 6.188561909s due to client-side throttling, not priority and fairness, request: POST:https://172.21.0.1:443/api/v1/persistentvolumes
```

Is there any configuration option for adjusting the qps/burst rate of the client? We've tried increasing `--worker-threads` but that doesn't seem to have helped much. It's not clear if the rate limit is per worker or is aggregated across workers.

PR #503, opened 2025-06-24
Allow overriding kubeclient burst and qps settings

badstreff:
We are running into rate limits and need a way to configure the client burst and qps settings.

In our use case, we spin up GitLab and GitHub jobs in their own pods. It's not uncommon for a job to kick off multiple downstream jobs, each of which needs a PVC. Once the client-side limit is reached, jobs begin queueing, resulting in a poor experience for our users.

This PR allows overriding these settings from the defaults.

It looks like we aren’t the only ones who need to adjust these settings — see #417.


derekbit (approved):
LGTM

PR #510, opened 2025-07-26
Change default nodePathMap value in storageClassConfigs in values.yml to []

Birkenstab:
Small fix in values.yml of chart

Because nodePathMap is an array and not an object


markusstephanides:
Was just about to create a PR with the same change but you were faster, thanks!

PR #522, opened 2025-09-02
add storageClass.allowedTopologies in helm chart

lizzzcai:
support `allowedTopologies` as we have dedicated node pool for local path storage.
example:
```yaml
  ## allowedTopologies field controls on which nodes the volumes can be dynamically provisioned, an empty value means no topology constraints
  allowedTopologies:
  - matchLabelExpressions:
    - key: example.com/local-ssd
      values:
      - "true"
```


lizzzcai:
Hi @derekbit , can you help to review this PR as well. This changes shouldn't affect any existing users.

PR #526, opened 2025-09-27
bump go 1.25

farazkhawaja:
Upgrade go version to go v1.25


rjourdan04:
> @derekbit is it worth cutting a new release so folks can start using this. This would have less CVEs

Hi, I would highly be interested in a new release to fix security issues !
Best regards,

PR #528, opened 2025-10-03
fix(chart): correct ServiceAccount namespace in ClusterRoleBinding

J3m3:
Update the clusterrolebinding.yaml to use
`{{ include "local-path-provisioner.namespace" . }}` so the ClusterRoleBinding correctly references the ServiceAccount in the intended Namespace when `namespaceOverride` is set.

The `namespaceOverride` feauture was added in the commit 772344504a9b ("feat: add ability to set custom namespace"), but the namespace of the ServiceAccount subject in the ClusterRoleBinding was not updated accordingly. As a result, it continued to use
`{{ .Release.Namespace }}`, causing permission errors when a custom namespace was specified.

### DIFF
--- a/deploy/chart/local-path-provisioner/templates/clusterrolebinding.yaml
+++ b/deploy/chart/local-path-provisioner/templates/clusterrolebinding.yaml
@@ -12,5 +12,5 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: {{ template "local-path-provisioner.serviceAccountName" . }}
-    namespace: {{ .Release.Namespace }}
+    namespace: {{ include "local-path-provisioner.namespace" . }}
 {{- end -}}


J3m3:
/cc @derekbit

Hi, I hope I CC’ed the right person, since it looks like you’re one of the most active maintainers. Could you please take a look at this PR when you have a chance? I didn’t find a detailed contribution guide, so please let me know if I should adjust anything. Thanks!

derekbit:
> /cc @derekbit
> 
> Hi, I hope I CC’ed the right person, since it looks like you’re one of the most active maintainers. Could you please take a look at this PR when you have a chance? I didn’t find a detailed contribution guide, so please let me know if I should adjust anything. Thanks!


No prob! @J3m3 I will check it soon. Sorry for late reply.

PR #530, opened 2025-10-29
fix: give clusterrole update on pvc

marcusramberg:
When provisioning fails, the controller tries to remove the selected node annotation to allow reprovisioning:
<https://github.com/rancher/local-path-provisioner/blob/58a077d589c04c95f1720a78b9618ebe8e2abea8/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v11/controller/controller.go#L1388>
But this fails with an RBAC error due to lack of permissions


derekbit (approved):
LGTM

derekbit (approved):

PR #542, opened 2025-12-24
fix: prohibit the reference path in pathPattern

mantissahz:
- enforce the prefix path in pathPattern
  The path in pathPattern should start with
  `{{ .PVC.Namespace }}/{{ .PVC.Name }}/`
- prohibit the reference path in pathPattern
  In a StorageClass, it should not be allowed:
  ```yaml
  ...
  parameters:
    pathPattern: '{{ .PVC.Namespace }}/{{ .PVC.Name }}/../etc/passwd'
  ...
  ```
  We should prohibit the reference path in the pathPattern for security issues.

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -317,7 +317,13 @@ func pathFromPattern(pattern string, opts pvController.ProvisionOptions) (string
 		return "", err
 	}
 
-	return buf.String(), nil
+	path := buf.String()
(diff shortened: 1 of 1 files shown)


derekbit (approved):

nlowe:
Note that this breaks existing installs if this condition was not met because this parameter cannot be changed after the storage class was created. I needed to `kubectl replace --force` the storage class to work around this.

PR #545, opened 2025-12-29
Grant helper pod only the priv it needs, explicitly set root bits

jcpunk:
This adds settings to the `helper-pod` so it has its security controls defined and setup explicitly.

With these in place it appears possible to run the main deployment as a fully non-root instance.


jcpunk:
I still believe this would be helpful

jcpunk:
I still believe this would be helpful

Issue #546, opened 2026-01-05
Prefix check on `pathPattern` breaking provisioning after upgrade: failed to create path from pattern: pathPattern must start with `{{ .PVC.Namespace }}/{{ .PVC.Name }}/`


utkuozdemir:
Hi,

I see that this change was recently introduced to enforce the `pathPattern` a prefix: https://github.com/rancher/local-path-provisioner/pull/542

I instead was using the pattern `{{ .PVC.Namespace }}_{{ .PVC.Name }}`.

As commented on the PR later, this change is not backwards compatible, as `StorageClass` spec is immutable after creation.

Furthermore, even if the storage class was replaced by `--force`, it creates an inconsistency in the filesystem between the paths generated by the old pattern (e.g., `myns_mypvc`) vs the ones generated with the new pattern (`myns/mypvc`).

I'd like to ask you to revisit that decision and consider reverting it, and doing the safety check in some other way than enforcing this prefix. If not, at least provide a flag to override this behavior, something like `allowUnsafePathPatterns`, so that I and possibly some other homelabbers can have a path forward.

utkuozdemir:
> Hi [@utkuozdemir](https://github.com/utkuozdemir), You can add an annotation `allowUnsafePathPattern: "true"` for the existing StorageClasses to skip the path check. ref: [#548](https://github.com/rancher/local-path-provisioner/pull/548)

Oh, it is an annotation, I see now. That was what I was missing.
For the helm chart, how does this look like: https://github.com/rancher/local-path-provisioner/pull/550

PR #547, opened 2026-01-06
fix: mitigate the impact of enforcing a pathPattern prefix

mantissahz:
ref: rancher/local-path-provisioner#546, https://github.com/rancher/local-path-provisioner/pull/542

### DIFF
--- a/provisioner.go
+++ b/provisioner.go
@@ -300,7 +300,7 @@ type pvMetadata struct {
 	PVC    metav1.ObjectMeta
 }
 
-func pathFromPattern(pattern string, opts pvController.ProvisionOptions) (string, error) {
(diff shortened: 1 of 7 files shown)


provisioner.go
```diff
@@ -377,13 +377,22 @@ func (p *LocalPathProvisioner) provisionFor(opts pvController.ProvisionOptions,
 
 	pathPattern, exists := opts.StorageClass.Parameters["pathPattern"]
 	if exists {
-		folderName, err = pathFromPattern(pathPattern, opts)
+		allowUnsafePath := false
+		allowUnsafePathPattern, exists := opts.StorageClass.Parameters["allowUnsafePathPattern"]
```
derekbit: Could you create a e2e test?

mantissahz: Added. Thanks.

PR #550, opened 2026-01-06
feat: allow specifying additional annotations on StorageClass

utkuozdemir:
Allow setting them via a Helm value.


derekbit (approved):
LGTM

PR #552, opened 2026-01-16
feat: allow setting securityContext on helper pods.

Coalmine-Canary:
# Changes
Small change to the helm chart to allow setting `securityContext` on the helper pods. 

# Why
In order for the helper pod to delete the contents of the PV, `runAsUser` or `priveleged` might be needed under the securityContext of the container.

# Testing
Tested on a bare metal cluster (using `runAsUser: 0` to allow deleting the contents of a PV that had been populated with data)


Coalmine-Canary:
Yeah it does, thanks I will close :)

PR #560, opened 2026-03-03
Support exposing user, group and perm patterns

### LINKED ISSUE #610: Support setting user/group/mode for created paths
It would be good if the PVC storage provisioned could specify the user, group and mode of the path created on disk.  Our primary use case is to allow local-path-provisioner to provision a PVC onto cluster filesystems where all nodes mount the filesystems and we can't allow root to own the files as the pod runs unprivileged and we need all the files owned by a user in LDAP.  In some cases we don't allow root to even write to our shared storage and in that case creating the PVC directory as unprivileged UID/GID would allow using the shared storage.

This ability to specify the UID/GID is going to be really useful for centers, like HPC sites, where shared storage is in use but there may not be Kubernetes storage plugins or those plugins have limitations.  Our current use case is GPFS and NFS with root_squash but this same pattern could easily be used with other shared filesystems like Lustre.

treydock:
This is still untested but the idea is to allow the setup command to be this:

```
install -d -u $VOL_USER -g $VOL_GROUP -m $VOL_PERM $VOL_DIR
```

We are overloading local-path-provisioner to allow PVC storage onto our cluster filesystems and having root:root and 0777 on a shared resource is far from ideal especially when we sometimes deal with very sensitive data on our clusters where a 0777 directory behind a 0700 directory still raises red flags.

Fixes #610

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,9 +237,9 @@ The scripts receive their input as environment variables:
 | `VOL_DIR` | Volume directory that should be created or removed. |
 | `VOL_MODE` | The PersistentVolume mode (`Block` or `Filesystem`). |
 | `VOL_SIZE_BYTES` | Requested volume size in bytes. |
-| `VOL_USER` | The user assigned from `userPattern` |
-| `VOL_GROUP` | The group assigned from `groupPattern` |
-| `VOL_PERM` | The permissions assigned from `permPattern` |
+| `VOL_USER` | User rendered from `userPattern`. |
+| `VOL_GROUP` | Group rendered from `groupPattern`. |
+| `VOL_PERM` | Permissions rendered from `permPattern`. |
 
 #### Reloading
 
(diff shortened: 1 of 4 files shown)


treydock:
@derekbit Could I get a review please?

treydock:
@innobead Could I get a review please?

derekbit:
> @derekbit Could I get a review please?

Sure. I will review it soon.

treydock:
@derekbit Could I get a review please?

derekbit:
> @derekbit Could I get a review please?

Sure. Sorry for the late review.

jcpunk:
I think this could still be useful.

treydock:
@derekbit Could this get a review please?

treydock:
@derekbit This could still use a review please.

derekbit:
@treydock Thanks for your contribution. Could you create a ticket in https://github.com/rancher/local-path-provisioner/issues?q=sort%3Aupdated-desc+is%3Aissue+state%3Aopen+ for explaning the purpose of the ticket and helping the issue management? Thanks.

treydock:
@derekbit I created #610 that hopefully explains what we're doing.  We've been using this PR for several months to provision PVCs onto our cluster filesystems and it has worked really well.

derekbit (approved):
LGTM

PR #582, opened 2026-06-03
chart: add optional SELinux node-setup DaemonSet

doccaz:
## Problem

On SELinux-enforcing nodes (e.g. RKE2 on RHEL/SUSE), the local-path-provisioner helper pods fail to create or delete volumes because the storage paths (e.g. `/opt/local-path-provisioner`) carry the wrong SELinux label. The helper pod's `container_t` process is denied write access by SELinux, so PVC provisioning silently fails.

The workaround today requires manual intervention on every node:

```bash
zypper -n in policycoreutils-python-utils
semanage fcontext -a -t container_file_t "/opt/local-path-provisioner(/.*)?"
restorecon -R -v /opt/local-path-provisioner/
```

This must be repeated for every new node that joins the cluster.

## Solution

Add an opt-in DaemonSet (`seLinuxNodeSetup.enabled=true`) that runs a privileged init container on every node to:

1. Create the storage path if it does not exist (`mkdir -p`).
2. Apply `chcon -Rt container_file_t` (or any configured SELinux type) to label it, allowing container processes to read and write it.

The DaemonSet tolerates all taints so it reaches every node, including control-plane nodes. The label is re-applied each time the pod starts, covering nodes that join the cluster after initial deployment.

The feature is **disabled by default** — no change for non-SELinux environments.

## Usage

```yaml
seLinuxNodeSetup:
  enabled: true
  paths:
    - /opt/local-path-provisioner   # match your nodePathMap paths
```

## Test plan

- [ ] Deploy chart with `seLinuxNodeSetup.enabled=false` (default) on a non-SELinux cluster — no DaemonSet created, no regression
- [ ] Deploy chart with `seLinuxNodeSetup.enabled=true` on an RKE2 cluster (RHEL/SUSE, SELinux enforcing) — DaemonSet runs on all nodes, labels the path, PVC provisioning succeeds without manual intervention
- [ ] Add a new node to the cluster — node-setup pod starts automatically and labels the path before any workload is scheduled

### DIFF
--- a/deploy/chart/local-path-provisioner/templates/selinux-node-setup.yaml
+++ b/deploy/chart/local-path-provisioner/templates/selinux-node-setup.yaml
@@ -1,0 +1,51 @@
+{{- if .Values.seLinuxNodeSetup.enabled }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "local-path-provisioner.fullname" . }}-node-setup
+  namespace: {{ include "local-path-provisioner.namespace" . }}
+  labels:
+{{ include "local-path-provisioner.labels" . | indent 4 }}
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "local-path-provisioner.name" . }}-node-setup
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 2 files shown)


jcpunk:
I thought I got the right default labeling into the upstream selinux policy....

Issue #208, opened 2021-09-03
request to provide health check probes (liveness/readiness) 


priyavj08:
currently no health checks like liveness/readiness probes available for the local-path-provisioner

please add these health checks to improve the reliability and provide a way to configure the timers through helm values

lukasz-bielinski:
Hi, do you have any estimates for adding probes?

jcpunk:
I'll confess interest in this as well

PR #583, opened 2026-06-09
Add initial /health probe

jcpunk:
This adds a trivial health probes per https://github.com/rancher/local-path-provisioner/issues/208


derekbit (approved):
LGTM

